PR #8, opened 2015-02-07
Faster (much faster) version of set_tensor_symmetry

jochym:
This is a much faster version of the set_tensor_symmetry_function from the harmonic/force_constants.py. It avoids explicit for loops and uses some numpy tricks for performance. The difference is substantial. For a large system (96at) it is few seconds vs. few minutes (or more). I have tested it against the old function - the results are identical. For large systems the old function was simply unusable.


atztogo:
Merged!

PR #11, opened 2016-01-06
Packaging for conda and small fix.

jochym:
The small fix is in siesta interface (__str__ method lacks self reference).
The packaging builds properly for python 2.7, 3.4 and 3.5


atztogo:
I merged to the develop branch. No serious conflict was found. Thank you.

PR #12, opened 2016-01-09
Add conda packaging info and fix python3 reversed iteration

jochym:
This patch fixes python 3 reverse iteration in continuous band plotting and adds conda packaging info to the package. Some of the packaging will need to be tweaked after merging this patch. I will fix any packaging problems after merging this patch and produce second patch to fix any problems.


atztogo:
Merged to develop branch. Thank for your fix.

PR #15, opened 2016-02-17
End VASP output files with newline and close file

brandongc:
Ran into an issue with a user here at NERSC using phonopy and our default VASP executable which is compiled with cray compilers. VASP compiled this way refuses to read POSCAR files unless every line ends with "\n."


brandongc:
Thanks, I patched this issue on our install so I don't think a new release is a critical issue for us. I'll keep an eye out for the next release though!

PR #16, opened 2016-02-25
Correct list divided by number in QHA

jochym:
In plotting routine for QHA there is a list divided by number (impossible operation).
I have added conversion to numpy array before division.
Works for me.
The other changes are already incorporated in your tree but the pull was somehow not properly merged so they are still in my history - please merge the whole pull request (all three changes) to make our histories consistent. Also please check if really only qha/__init__.py is changed before committing - if not reject the patch and let me know, and I will redo the pull request.


atztogo:
Merged to the develop branch. Thanks.

PR #20, opened 2016-05-12
Packaging for both pypi and conda

jochym:
Do not be scared by large number of commits. It takes some tests to figure all this up. Actual changes are not that large (see the diffs). The problem with unfolding shows that we should move this whole CI business to your tree. It will help to catch problems during development instead of packaging.


atztogo:
I merged it to develop branch. Thank you!

PR #24, opened 2016-09-15
minor bug fix in the printing of irreducible rep label

matk86:
* string concatenation fails if the label is None


atztogo:
I think this is already fixed in develop branch in a similar way.
https://github.com/atztogo/phonopy/blob/develop/phonopy/phonon/irreps.py#L1163
Could your review the develop branch? Thanks!

matk86:
yes, its fixed in the develop branch. thanks

PR #25, opened 2016-09-16
bug fix in Atoms

matk86:
* fix the set_cell api in Atoms: updating cell doesnt update the scaled_positions
*  some minor pep8 formatting fixes in the atoms.py module


atztogo:
Thanks. I'm going to merge.

For irreps, I will prepare simple tests and check if your commit works well. It will take some time since I have to find phonon results having those point groups.

On Atoms, I could not understand your comment
- fix the set_cell api in Atoms: updating cell doesnt update the scaled_positions
I think scaled_positions is not necessary to be updated by updating cell. What is the actual problem that you have met?
In fact, I don't like the Atoms class and want to replace it by something like that shown below (tuple one),
http://atztogo.github.io/spglib/python-spglib.html#crystal-structure-cell

matk86:
The issue was that the atoms class let's one initialize the object with out the cell and/or positions and if you try to set the cell later using the set_cell API it doesn't complain about not having the positions, causing other things to fail later ......I was trying to ensure that setting the cell and positions go hand in hand..I will take a look at your suggestion and try to port it to phonopy.. thanks

atztogo:
For Atoms, I think you don't need to port that I suggested (I didn't mean to ask you to do it). Instead, I will change the code to raise an exception when any of three values, cell, scaled_positions (=positions), and numbers (=symbols), are missing for set_cell, etc.

atztogo:
I merged all your changes to develop and then
- I found a problem on -43m database and fixed it. But does this work for you?
- I updated Atoms class so as to raise exception when not having positions or for similar cases, i.e., some consistency check is now done. Please check it if you like it or not.

matk86:
thank you!
I am not sure whether you are planning to do it or not but i think it would be nice to add the basis functions to the character dictionary so that we can predict whether the phonon mode is raman or ir active..please let me know your thoughts on it...i can help with it if you are too busy..
thanks for all your help.

-Kiran Mathew

atztogo:
Currently I have no plan to do it. But I think there is no problem for phonopy to include it. For it, the database should be separated from irreps.py and moved to some other place since it became too many lines, which prevents the readability of the logic part.

Once the irrep is known, we can easily refer other source of such information, e.g., http://www.cryst.ehu.es/ . A problem of such a database is that internet connection is mandatory. If a computer accessible database that doesn't need internet connection is expected and if such a database doesn't exist , I think it's a better idea to create a new project focusing on it, in such a way, it is separately maintained.

PR #26, opened 2016-09-19
load the character table from json file.

matk86:
This is a follow up to our conversation on my previous pull request(#25). 

I agree  with you on the fact that adding all the character table info into the irrep.py module will make the code unreadable, but i don't think we need a database to store the character table information, all the data could be loaded from a json file. 
In this pull request i have moved the character table information to 'charcater_table.json' file and then loaded in the irreps.py module. I have run the irreps unittests in the test folder and they all run fine in python 2 and 3. Apart from that, I have also added an extra key called "functions" to store the linear and quadratic function information. At the moment i have added the new key only to a few point groups, if you like these changes i can go through the rest of the point groups too. 

Example:
 the entry in the table for group 4/m now looks like:
{'character_table': {'Ag': [1, 1, 1, 1, 1, 1],
                     'Au': [1, 1, 1, -1, -1, -1],
                     'Bg': [1, -1, 1, 1, -1, 1],
                     'Bu': [1, -1, 1, -1, 1, -1],
                     'Eg': [2, 0, -2, 2, 0, -2],
                     'Eu': [2, 0, -2, -2, 0, 2]},
 'functions': {'linear': [['Rz'], [], [['Ry', 'Rx']], ['z'], [], [['x', 'y']]],
               'quadratic': [['x2+y2', 'z2'],
                             ['x2-y2', 'xy'],
                             [['xz', 'yz']],
                             [],
                             [],
                             []]},
 'mapping_table': {'C2': [[[-1, 0, 0], [0, -1, 0], [0, 0, 1]]],
                   'C4': [[[0, -1, 0], [1, 0, 0], [0, 0, 1]],
                          [[0, 1, 0], [-1, 0, 0], [0, 0, 1]]],
                   'E': [[[1, 0, 0], [0, 1, 0], [0, 0, 1]]],
                   'S4': [[[0, 1, 0], [-1, 0, 0], [0, 0, -1]],
                          [[0, -1, 0], [1, 0, 0], [0, 0, -1]]],
                   'i': [[[-1, 0, 0], [0, -1, 0], [0, 0, -1]]],
                   'sgh': [[[1, 0, 0], [0, 1, 0], [0, 0, -1]]]},
 'rotation_list': ['E', 'C4', 'C2', 'i', 'S4', 'sgh']}


atztogo:
If 32 crystallographic point groups will be completed, additional keywords will not be so painful even if some bugs are found by chance. Let's forget about how to show those new data here now. Other than these points, I have two comments below.

1. To prepare the database in widely accepted format is a good idea. However for phonopy, I think there is no reason to read the database from a json file. So for example, I like that this is solved by preparing a text converter from json data to python dictionary code and the python code is stored in the code tree as phonopy/structure/pointgroup_irreps.py and the json data and converter are kept only in the git repository. The json file may be maintained in a different way (e.g. by you) if phonopy can use it in this way in the point of view of the license. License issue is always tricky.

2. For the maintenance reason (it seems I have to keep supporting phonopy for many years...), it is better for me it is written in a human readable way. This is the reason that I write the database in irreps.py against pep8.

3. If the point group database becomes too nice, I'm a bit afraid that in the future somebody expects those character tables for non-gamma points ;-)

matk86:
Ok

PR #28, opened 2016-11-19
Make tests discoverable and actually trigger on errors

jochym:
This is a slight rework of test suite. Modifications uncovered some tests which were failing and ignored.
Now all runs from `python -m unittest discover` and actually triggers when there are errors present.
The phonopy/symmetry.py is adapted for changes in spglib get_symmetry api (no use_magmom arg).
There are 4 failing tests in the set. Two of them look minor, but two are serious (symmetry tests).
@atztogo I know too little about internals of spglib/phonopy to correct this. Please, review the changes. I think that the breaks should be investigated - it may be quite serious.


atztogo:
Thanks for your a lot of work. I have fixed the test failures, and now tests pass.

jochym:
Great. Happy to help. If we have tests why not use them ;) I have noticed that you actually spotted some problems due to this work. Good. Are you planning a release soon?

atztogo:
Yes, I will release but not changing the version number, i.e, I just want to merge these changes to master branch. However by merging in the rc branch, travis failed in the deployment step. Currently I am checking why it happened. I see you haven't change .travis.yaml except for harmless lines...

atztogo:
Now travis went well. I found "cd ../.." is lost in script section. Thanks.

jochym:
Good - please read the issue on spglib under windows I just submitted. Maybe you can do something about it. I do not know why the compilation fails there.

PR #34, opened 2016-12-07
phonophy/interface/FHI-aims.py: Bugfix (read of forces)

mospring:
FHI-aims output of forces for atoms is
| 999	force1	force2	force3
for less than 1000 atoms and
|1000	force1	force2	force3
for more than 999 atoms. Therefore column have to be counted in reversed order.

### DIFF
--- a/phonopy/interface/FHIaims.py
+++ b/phonopy/interface/FHIaims.py
@@ -161,7 +161,7 @@ def read_aims_output(filename):
             forces = []
             for i in range(N):
                 l += 1
-                force = map(float, lines[l].split()[2:5])
+                force = map(float, lines[l].split()[-3:])
                 forces.append(force)
         l += 1
     


atztogo:
I am not able to review this pull request. This is because I am not a user of FHI-aims. FHI-aims.py  and phonopy-FHI-aims were written by people in FHI very before. It's possible just accepting this request but I am not sure if I should do it or not. I would like somebody in FHI to correspond to FHI-aims.py as a maintainer, but I don't know who will be. Do you know such a person?

Now phonopy-FHI-aims is too old to maintain and to support many useful features in phonopy. Currently phonopy has a simple system (https://github.com/atztogo/phonopy/blob/master/phonopy/interface/__init__.py) to treat different force calculators such as VASP, QE, abinit. I recommend  FHI-aims-phonopy users to use normal phonopy interface, for which I have to write some code. The issue for me is the testing. To do it, clearly I need help from FHI-aims users. Do you have any idea for that?

PR #41, opened 2017-04-07
Optimize `get_equivalent_smallest_vectors` (also, bugfix for HDF5 in python3)

ExpHP:
This optimizes `get_equivalent_smallest_vectors` as [noted here](https://github.com/atztogo/phonopy/issues/40).

It also contains a fix for the `--hdf5` flag in python3, which currently throws an exception.

---

### Info on benchmarks

I tried out various strategies for optimizing this function. The following table collects benchmarks for modifications to the function.  Most were not worth it; see [this table for details](https://github.com/ExpHP/phonopy-eigenvector-tblg-benchmark/edit/master/info/benchmarks.md).

_**This PR only contains the two optimizations found to have the biggest bang for the buck.**_

---

### Further possible optimizations

The big remaining bits I see that can be eliminated now are the delaunay reduction and matrix inverses. **These should probably be precomputed and placed as members on some class, or passed in as an argument.** Unfortunately I don't know the code well enough to know where they best belong.

I experiment with this in the final row in the table (I bench a modification where `get_equivalent_smallest_vectors` takes `reduced_bases` as an argument, which produces another decent 50% speedup).

### DIFF
--- a/phonopy/file_IO.py
+++ b/phonopy/file_IO.py
@@ -191,9 +191,7 @@ def parse_FORCE_CONSTANTS(filename="FORCE_CONSTANTS"):
 def read_force_constants_hdf5(filename="force_constants.hdf5"):
     import h5py
     with h5py.File(filename, 'r') as f:
-        fc = f[f.keys()[0]][:]
-        return fc
-    return None
+        return f[next(iter(f.keys()))][:]
 
 #
 # disp.yaml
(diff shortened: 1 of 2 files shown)


atztogo:
Merged to the master branch which invokes its PyPI and conda releases. Thanks for your beautiful hack, which is fast and readable. Thanks also for the detailed explanation. I replaced the line np.linalg.norm(x, axis=1) by np.sqrt(x**2.sum(axis=1)) because older numpy doesn't support axis keyword in np.linalg.norm.

atztogo:
A follow-up. I moved get_smallest_vectors method from harmonic/dynamical_matrix.py to structure/cells.py, and it is changed to be called only once at the initialization of Primitive class object for the usual phonopy run. These data are stored in each Primitive class object and accessed by primitive.get_smallest_vectors().
https://github.com/atztogo/phonopy/blob/master/phonopy/structure/cells.py#L299

Togo

PR #44, opened 2017-06-23
Info on conda-forge packages and few small stylistic changes

jochym:
I have added info about conda-forge packages to the install page.
Conda-forge seems to be very popular (almost 10000 downloads of phonopy from their channel!) and it provides builds for Mac and Windows computers as well as full matrix of linux builds.
I think we should inform the users about it. In addition for  windows and Mac users this is by far the simplest path.


atztogo:
Thanks. I updated the document web site. Can you announce it in the mailing list or should I do it?

jochym:
Do it please. I am maintaining spglib/phonopy/ase packages on conda-forge
for some time already. They have downloads in high thousands already so
seem fairly popular.

P.

pt., 23 cze 2017 o 14:12 użytkownik Atsushi Togo <notifications@github.com>
napisał:

PR #46, opened 2017-08-12
Fix O(n^3) scaling in get_smallest_vectors

ExpHP:
This is another optimization of `get_smallest_vectors`.

Most of the logic of `get_equivalent_smallest_vectors` is now inlined into `get_smallest_vectors` where it can be lifted outside of the loop as opposed to being recomputed O(n^2) times.  This is important considering that some of that logic involved computing data for all N positions, even though it only needed two of them! A new function `get_equivalent_smallest_vectors_simple` now holds the simpler, core functionality of that function.

`get_equivalent_smallest_vectors` is no longer used by `get_smallest_vectors`, but it still exists because other code uses it.

### DIFF
--- a/phonopy/structure/cells.py
+++ b/phonopy/structure/cells.py
@@ -457,35 +457,53 @@ def get_equivalent_smallest_vectors(atom_number_supercell,
                                     primitive_lattice,
                                     symprec):
     reduced_bases = get_reduced_bases(supercell.get_cell(), symprec)
-    positions = np.dot(supercell.get_positions(), np.linalg.inv(reduced_bases))
+    reduced_bases_inv = np.linalg.inv(reduced_bases)
+    cart_positions = supercell.get_positions()
 
-    # Atomic positions are confined into the lattice made of reduced bases.
-    positions -= np.rint(positions)
+    # Atomic positions are confined into the delaunay-reduced superlattice.
+    # Their positions will lie in the range -0.5 < x < 0.5, so that vectors
+    # drawn between them have components in the range -1 < x < 1.
+    def reduced_frac_pos(i):
+        vec = np.dot(cart_positions[i], reduced_bases_inv)
+        return vec - np.rint(vec)
+    p_pos = reduced_frac_pos(atom_number_primitive)
+    s_pos = reduced_frac_pos(atom_number_supercell)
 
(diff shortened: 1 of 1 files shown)


phonopy/structure/cells.py
```diff
@@ -457,35 +457,53 @@ def get_equivalent_smallest_vectors(atom_number_supercell,
...
-    p_pos = positions[atom_number_primitive]
-    s_pos = positions[atom_number_supercell]
-
-    # The vector arrow is from the atom in primitive to
-    # the atom in supercell cell plus a supercell lattice
-    # point. This is related to determine the phase
-    # convension when building dynamical matrix.
-    supercell_vectors = np.array([
+    reduced_bases_inv = np.linalg.inv(reduced_bases)
+    cart_positions = supercell.get_positions()
```
ExpHP: oh, oops, this still does O(n) work.  I meant to replace this with `get_scaled_positions` and then only multiply the positions we need by the cell.

In any case, I think I'll leave it for now since I don't use any of the functionality that calls this function.

atztogo:
Merged to the master branch. Thanks @ExpHP . 
I eliminated use of `get_equivalent_smallest_vectors` from phonopy. But I keep it since it is still used in phono3py.
For `get_smallest_vectors` and `get_equivalent_smallest_vectors_simple`, I put underscore in front of these names because they are local method in this file, though I'm not sure if this is a good convention or not.
The other changes are just made to remove unnecessary codes that were used for keeping the compatibility with older phonopy.

phonopy/structure/cells.py
atztogo: You don't need to think about improving the computational complexity of `get_equivalent_smallest_vectors`. It is still used in phono3py, but phono3py is supposed for smaller supercell cases only due to its heavy computational demand in different part. But in fact, I thought so as well for phonopy... People easily go beyond my imagination.

ExpHP:
>  For get_smallest_vectors and get_equivalent_smallest_vectors_simple, I put underscore in front of these names because they are local method in this file, though I'm not sure if this is a good convention or not.

Yes, this is a good convention, and I was glad to see that change get added. :)

Indeed, it's always good to keep your public API small whenever possible, because functions are often written with assumptions in mind that might not be clear from the outside.

Issue #36, opened 2016-12-22
disp.yaml wrong symbol and mass


baranas:
Hi, Togo

I am using your software for defect vibrational spectrum calculations (pwscf and vasp).
For my calculations I am using your FORCE_CONSTANTS file and displ.yaml. 
It is awesome :). On pwscf it works fine, however for vasp displ.yaml it mixes atomic 
symbols and masses.

For diamond
POTCAR

    PAW_PBE C 08Apr2002
    4.00000000000000000 
    ...

POSCAR

```
diamond 
    3.5620940727273465
      1.00 0.00 0.00
      0.00 1.00 0.00
      0.00 0.00 1.00
8
Direct
    0.000000000    0.000000000    0.000000000
    0.500000000    0.500000000    0.000000000
    0.000000000    0.500000000    0.500000000
    0.500000000    0.000000000    0.500000000
    0.250000000    0.250000000    0.250000000
    0.750000000    0.750000000    0.250000000
    0.750000000    0.250000000    0.750000000
    0.250000000    0.750000000    0.750000000    
```

It generates disp.yaml

```
natom:  216
displacements:
- atom:    1
  direction:
    [   1.0000000000000000,  0.0000000000000000,  0.0000000000000000 ]
  displacement:
    [   0.0100000000000000,  0.0000000000000000,  0.0000000000000000 ]
- atom:    1
  direction:
    [  -1.0000000000000000,  0.0000000000000000,  0.0000000000000000 ]
  displacement:
    [  -0.0100000000000000,  0.0000000000000000,  0.0000000000000000 ]
lattice:
- [    10.686282218182040,     0.000000000000000,     0.000000000000000 ] # a
- [     0.000000000000000,    10.686282218182040,     0.000000000000000 ] # b
- [     0.000000000000000,     0.000000000000000,    10.686282218182040 ] # c
points:
- symbol: H  # 1
  coordinates: [  0.000000000000000,  0.000000000000000,  0.000000000000000 ]
  mass: 1.007940
- symbol: H  # 2
  coordinates: [  0.333333333333333,  0.000000000000000,  0.000000000000000 ]
  mass: 1.007940                                 
...
```

with symbols H and masses 1.007940. I can overcome this in my software,
but in future it would be nice if you could fix that. By the way is it possible to contribute
to your code, I would like to optimise fc_symmetry which is very usefull for me.
Again thanks for such nice and usefull package.

atztogo:
[...]

PR #47, opened 2017-08-13
add compute_permutation C function

ExpHP:
This adds `_compute_permutation` which is an implementation of the overlap searches in `distribute_fc2`, but with performance characteristics that scale better for very large structures.

The key to its speed [actually lies here, in the python bits](https://github.com/ExpHP/phonopy/blob/eigenvector-hack/phonopy/harmonic/force_constants.py#L683-L693): Sorting is used to precondition the data into a form that lets `compute_permutation` work fast.

**NOTICE: In this commit, `distribute_fc2` still does not yet use the new function!!**  That is because I haven't yet figured out how to incorporate all of your recent changes.

FYI, here's how I used to use it, back before you added `distribute_fc2_all`:

* I used it to [break `_distribute_fc2_part` into two steps](https://github.com/ExpHP/phonopy/blob/eigenvector-hack/phonopy/harmonic/force_constants.py#L647-L671)
* The second step was [just this simple C function](https://github.com/ExpHP/phonopy/blob/eigenvector-hack/c/_phonopy.c#L690-L716)

---

However, we still get big wins in this commit, because I was able to use it in `get_positions_sent_by_rot_inv` instead.  On [these input files](https://github.com/ExpHP/phonopy-eigenvector-tblg-benchmark/tree/master/input/big), this change speeds that function up from taking ~50s (out of ~250s total runtime for `--readfc --band=...`) down to only taking ~0.12s.

### DIFF
--- a/c/_phonopy.c
+++ b/c/_phonopy.c
@@ -52,6 +52,7 @@ static PyObject * py_get_derivative_dynmat(PyObject *self, PyObject *args);
 static PyObject * py_get_thermal_properties(PyObject *self, PyObject *args);
 static PyObject * py_distribute_fc2(PyObject *self, PyObject *args);
 static PyObject * py_distribute_fc2_all(PyObject *self, PyObject *args);
+static PyObject * py_compute_permutation(PyObject *self, PyObject *args);
 
 static int distribute_fc2(double *fc2,
 			  PHPYCONST double lat[3][3],
@@ -63,6 +64,14 @@ static int distribute_fc2(double *fc2,
 			  PHPYCONST int r[3][3],
 			  const double t[3],
 			  const double symprec);
+
+static int compute_permutation(int * rot_atom,
+				  PHPYCONST double lat[3][3],
+				  PHPYCONST double (*pos)[3],
+				  PHPYCONST double (*rot_pos)[3],
+				  const int num_pos,
+				  const double symprec);
+
 static PyObject * py_thm_neighboring_grid_points(PyObject *self, PyObject *args);
 static PyObject *
 py_thm_relative_grid_address(PyObject *self, PyObject *args);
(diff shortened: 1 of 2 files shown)


atztogo:
Merged to the master branch. Thanks.

I've not yet read this pull request very much.

I modified slightly to avoid using symprec**2. I don't remember clearly, but I had some problem before, see https://github.com/atztogo/spglib/issues/36. Can this change is acceptable to you? I ask because sqrt(distance2) is a bit heavy computation.

ExpHP:
Hm.  I personally can't picture how such a problem would arise in this case unless somebody is specifying something like `SYMPREC = 1e-150` (machine epsilon is irrelevant since we aren't e.g. adding these numbers or subtracting them from a constant), but then again I don't understand all the details behind #36.

If you'd rather be safe, it's probably not as big a deal as it seems performance-wise.  I've heard sqrts are actually pretty fast on x86_64.  I'll benchmark it later when I get the chance

ExpHP:
> I've not yet read this pull request very much.

This concerns me a bit since you've already merged it into master. (very quickly, too)  I hope you at least have a good idea of what the changes are doing?

Please don't feel rushed to merge a PR, especially when it affects an important part of the code like the force constants.  You should take time to ask questions, and don't merge it until you feel comfortable maintaining it as part of your code base.  I can wait! :)

atztogo:
You are very correct.

Maybe I get to remember why it happened for spglib. In spglib, something like SYMPREC = 1e-150 can happen due to its iterative algorithm where SYMPREC was systematically reduced until the condition met, and in some bad case, the condition never meets. This algorithm itself could be bad. But at least it was safe not to use SYMPREC**2 in this routine. After that, I tend to avoid SYMPREC**2 for any case. Implementation of spglib is quite tough and became complicated to tolerate distortion of crystal structure. I need charm to rely on :)

So here it's not important to avoid using SYMPREC**2, but if it doesn't impact so much for performance, maybe I'd like to keep my local convention.

atztogo:
On merging to master, these days I don't take it so serious. Your PR passed the tests, which is basically enough. If I postpone it, I may not find a chance to merge it for a while, which is more scary for me. I will soon read your PR during my commutating between home and university. I would appreciate if you could continue your development.

ExpHP:
Okay, I see where you are coming from.  Yeah, it certainly can be easy for PRs to get left in limbo.

Thanks for the sentiment, and cheers!

PR #48, opened 2017-08-13
use `_compute_permutation` in `distribute_fc2`

ExpHP:
NOTE: **I have not tested the case where `atom_list != range(num_pos)`** since it doesn't seem to me that I can easily test it using input files.  Maybe there should be a unit test that calls `phonon.produce_force_constants(calculate_full_force_constants=False)`?
Please let me know if this is an issue.

(Just as an aside, my main method of testing my changes are to do bit by bit comparisons of band.yaml.  I make sure to preserve the order of all meaningful floating point operations so that the file produced in the end is identical; I do this simply because it makes it easier to find bugs)

---

Here's some numbers, doing `--writefc` on [these input files](https://github.com/ExpHP/phonopy-eigenvector-tblg-benchmark/tree/master/input/big):
- **before:**
  - `total runtime`: 217.664s
  - `distribute_force_constants`: **63.720s** (30% of total)
- **after:**
  - `total runtime`: 162.016
  - `distribute_force_constants`: **0.491s** *(negligible!)*

### DIFF
--- a/c/_phonopy.c
+++ b/c/_phonopy.c
@@ -52,6 +52,7 @@ static PyObject * py_get_derivative_dynmat(PyObject *self, PyObject *args);
 static PyObject * py_get_thermal_properties(PyObject *self, PyObject *args);
 static PyObject * py_distribute_fc2(PyObject *self, PyObject *args);
 static PyObject * py_distribute_fc2_all(PyObject *self, PyObject *args);
+static PyObject * py_distribute_fc2_with_mappings(PyObject *self, PyObject *args);
 static PyObject * py_compute_permutation(PyObject *self, PyObject *args);
 
 static int distribute_fc2(double *fc2,
@@ -64,6 +65,16 @@ static int distribute_fc2(double *fc2,
 			  PHPYCONST int r[3][3],
 			  const double t[3],
 			  const double symprec);
+
+static void distribute_fc2_with_mappings(double (*fc2)[3][3],
+                                         const int * atom_list,
+                                         const int len_atom_list,
+                                         PHPYCONST double (*r_carts)[3][3],
+                                         const int * permutations,
+                                         const int * map_atoms,
+                                         const int * map_syms,
+                                         const int num_rot,
+                                         const int num_pos);
 
 static int compute_permutation(int * rot_atom,
 				  PHPYCONST double lat[3][3],
@@ -128,6 +139,8 @@ static PyMethodDef _phonopy_methods[] = {
   {"distribute_fc2", py_distribute_fc2, METH_VARARGS, "Distribute force constants"},
   {"distribute_fc2_all", py_distribute_fc2_all, METH_VARARGS,
    "Distribute force constants for all atoms in atom_list"},
+  {"distribute_fc2_with_mappings", py_distribute_fc2_with_mappings, METH_VARARGS,
+   "Distribute force constants for all atoms in atom_list using precomputed symmetry mappings."},
   {"compute_permutation", py_compute_permutation, METH_VARARGS,
    "Compute indices of original points in a set of rotated points."},
   {"neighboring_grid_points", py_thm_neighboring_grid_points,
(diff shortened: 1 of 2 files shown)


phonopy/harmonic/force_constants.py
```diff
@@ -894,3 +935,45 @@ def permutation_error():
...
+#  * map_atoms, shape [num_pos].
+#    Maps each atom in the full structure to its equivalent atom in atom_list_done.
+#    (each entry will be an integer found in atom_list_done)
+#
+#  * map_syms, shape [num_pos].
+#    For each atom, provides the index of a rotation that maps it into atom_list_done.
+#    (there might be more than one such rotation, but only one will be returned)
+#    (each entry will be an integer 0 <= i < num_rot)
+def _get_sym_mappings_from_permutations(permutations,
+                                        atom_list_done):
```
ExpHP: this function probably isn't necessary.  I added it because I had big plans on how it could be optimized, but after benchmarking I found that it is plenty fast already, even as a brute force loop!

I don't know how much faster it is than e.g. get_atom_mapping_by_symmetry, but I doubt it is by very much.

ExpHP:
Oops, need to fix my declarations to be pre-C99 compatible

c/_phonopy.c
```diff
@@ -65,6 +66,16 @@ static int distribute_fc2(double *fc2,
...
 
+static void distribute_fc2_with_mappings(double (*fc2)[3][3],
+                                         const int * atom_list,
+                                         const int len_atom_list,
+                                         PHPYCONST double (*r_carts)[3][3],
+                                         const int * permutations,
+                                         const int * map_atoms,
+                                         const int * map_syms,
+                                         const int num_rot,
+                                         const int num_pos);
```
ExpHP: (meant to post this as a regular comment, not a review)

this function signature bothers me because map_atoms and map_syms are basically implementation details.

we could take my _get_sym_mappings_from_permutations function (at the bottom of the diff) and convert it into a C helper function, so that this function could then take atom_list_done instead of map_atoms and map_syms. I did try to do this at some point, but was discouraged by how gnarly even the simple logic in _get_sym_mappings_from_permutations becomes when written in C

atztogo:
> Maybe there should be a unit test that calls `phonon.produce_force_constants(calculate_full_force_constants=False)`?

No. That was made before I realized the existence of unittest. I am an old programmer. But I think it is trivial.

At the first step, force constants are calculated between atoms in primitive cell and atoms in supercell. Therefore the force constants matrix is unnecessary to be filled fully due to translational symmetry. 'fully' means those elements for atoms between supercells. However by numerical noise, the translational symmetry can be broken, and in phonopy there is an option to force the translational symmetry, which requires 'full' force constants matrix. When using a good force calculator, the degree of broken symmetry is negligible often, so we need the former force constants, which corresponds to `phonon.produce_force_constants(calculate_full_force_constants=False)`. 

Therefore if we test it, what has to be tested is whether the set of atomic indices contains all atomic indices of atoms in primitive cell or not.

atztogo:
>Oops, need to fix my declarations to be pre-C99 compatible

Yes, unfortunately this is necessary to support old Microsoft Visual C(++), which is still used to make conda package with python-2.7.

atztogo:
Your performance gain of `distribute_force_constants` is considerable. Recent my update is just to improving performance using openMP concurrency. Because your algorithm is quite fast, so my recent update can be ignored.

atztogo:
I don't know well how to use github conversation, so I write it an ugly way below.

https://github.com/atztogo/phonopy/blob/ce40b082bfa6e3a3c7b22dcd48d8abd2e4d89a31/phonopy/harmonic/force_constants.py#L834
```
    def sort_by_lattice_distance(fracs):
        carts = np.dot(fracs - np.rint(fracs), lattice.T)
        perm = np.argsort(np.sum(carts**2, axis=1))
        sorted_fracs = fracs[perm]
        return perm, sorted_fracs
```

This `sorted_fracs` is passed to c code. So I think it's better to write

```
        sorted_fracs = np.array(fracs[perm], dtype='intc', order='C')
```

This is a kind of charm. Sometimes numpy just resorting the index order only but not for data in memory. If I am sure this is not the case always, I don't need to do this. `order='C'` is also redundant, since it seems it is the default parameter. But for multidimensional numpy array, I do this for sure to make the data in memory correct order. The numpy behaviour may depend on its version, too, which I am afraid of. But as far as I do this, I didn't have troubles.

Related to this, I have a small comment on my habit.

https://github.com/ExpHP/phonopy/blob/ba95ab7a003ce13874aee25bf250fc54865559a8/phonopy/harmonic/force_constants.py#L842
```
out = []
```
If the array size of out is known before the following loop. I prefer to declare the array before
```
out = np.zeros((nsym, npos), dtype='intc', order='C')
```
to help the readability and to make me relax to expect collect memory order.

atztogo:
`//` in C- code comments is not allowed for old C style compatibility, unfortunately...

atztogo:
Overall, I am ready to merge.
1. Do you want to remove `_get_sym_mappings_from_permutations` before merging or after?
2. Can I modify the numpy array declarations that I wrote above?

By the way, do you hate pointer arithmetic? I grew up with old computers and some machine language, touching memory directly is very usual, but not a good convention.

ExpHP:
> This is a kind of charm. Sometimes numpy just resorting the index order only but not for data in memory. If I am sure this is not the case always, I don't need to do this. order='C' is also redundant, since it seems it is the default parameter. But for multidimensional numpy array, I do this for sure to make the data in memory correct order.

Oops! Yes, I forgot to consider this, but yes, this is very important here since the C code uses PyArray_DATA directly.  (in other projects I've had the C code call `PyArray_ContiguousFromAny`, but that only solves part of the problem; it's still possible to forget other important things like `dtype=intc`)

>  Do you want to remove _get_sym_mappings_from_permutations before merging or after?

I'll work on it after.

> Can I modify the numpy array declarations that I wrote above?

Of course!

ExpHP:
> If the array size of out is known before the following loop. I prefer to declare the array before to help the readability and to make me relax to expect collect memory order.

Thanks for explaining this reasoning.  I'll try to keep this in mind in the future.

Really, the only reason for my habit of using list and `append` is to avoid having another index variable (since I'm afraid I might make a typo and use the wrong index somewhere).

> By the way, do you hate pointer arithmetic? I grew up with old computers and some machine language, touching memory directly is very usual, but not a good convention.

Heh, what makes you say that?  I definitely don't enjoy it, since there's a lot more room to make mistakes when you have to manually write out strides and offsets all the time. (e.g. an easy accident is to write `i * num_rot + j * 9` when you meant `i * num_rot * 9 + j * 9`).  That said, I'm not exactly alien to it, either. :)

atztogo:
Merged to the master branch. Your code is much well written by me. I learn computer programing little by little from you. Thanks.

On pointer arithmetic, you are correct. I use a lot of it in phono3py. I have to be careful not to make mistake in strides. What I would mean was about this line,
```
permutation = &permutations[sym_index * num_pos];
```
I normally write
```
permutation = permutations + sym_index * num_pos;
```
As you write in the last line, I have to remember that the point arithmetic is type-dependent with respect to real (though virtual) memory address.

ExpHP:
> I don't know well how to use github conversation, so I write it an ugly way below.

Some tips: The way I do those comments on the code lines is that I go to the [files tab](https://github.com/atztogo/phonopy/pull/48/files) and click on a line of code; it will let you write a comment.  This feature is often useful for requesting changes to PRs.

Also, if you write your code blocks using \`\`\`python or \`\`\`c instead of by indenting 4 spaces, you get syntax highlighting.

```
**indented 4 spaces**

    def f(x):
        pass

**better highlighting** (replace ' by `)

'''python
def f(x):
    pass
'''
```

**indented 4 spaces**

    def f(x):
        pass

**better highlighting**

```python
def f(x):
    pass
```

atztogo:
Thanks!

> I go to the Changes tab and click on a line of code

What a nice feature!

atztogo:
Follow-up. distribute_fc2 and distribute_fc2_all are removed. We will maintain only distribute_fc2_with_mappings.

PR #50, opened 2017-08-15
optimize get_smallest_vectors once more with numpy broadcasting and C

ExpHP:
Despite the algorithmic improvements in recent commits, `get_smallest_vectors` is once again back at the top of the profiler results.  Unfortunately, I was not able to come up with any more algorithmic improvements, so I instead had to resort to careful use of numpy features to maximize data-crunching potential.

One last bit was difficult to implement using features available in numpy, so I wrote it in C instead.

---

On [these input files](https://github.com/ExpHP/phonopy-eigenvector-tblg-benchmark/tree/605f0d662fa8/input/big), using `--readfc`:
- **before:**
  - `total runtime`: 138.351s
  - `_get_smallest_vectors`: **73.126s** (53% of total)
- **after:**
  - `total runtime`: 71.632s
  - `_get_smallest_vectors`: **5.642s** (8% of total)

For reference, `eigh` (the "real work") takes 20.794s on these inputs.

### DIFF
--- a/phonopy/structure/cells.py
+++ b/phonopy/structure/cells.py
@@ -528,40 +528,75 @@ def _get_smallest_vectors(supercell, primitive, symprec):
       [atom_super, atom_primitive]
     """
 
+    # useful data from arguments
     p2s_map = primitive.get_primitive_to_supercell_map()
     size_super = supercell.get_number_of_atoms()
     size_prim = primitive.get_number_of_atoms()
-    shortest_vectors = np.zeros((size_super, size_prim, 27, 3), dtype='double')
-    multiplicity = np.zeros((size_super, size_prim), dtype='intc')
-
     reduced_bases = get_reduced_bases(supercell.get_cell(), symprec)
-    reduced_bases_inv = np.linalg.inv(reduced_bases)
-    primitive_lattice = primitive.get_cell()
-    primitive_lattice_inv = np.linalg.inv(primitive_lattice)
 
-    # matrix that converts fractional positions in the reduced bases into
-    #  fractional positions in the primitive lattice
-    supercell_to_primitive_frac = reduced_bases.dot(primitive_lattice_inv)
-
-    # all positions are reduced into the cell formed by the reduced bases
-    supercell_fracs = np.dot(supercell.get_positions(), reduced_bases_inv)
+    # Reduce all positions into the cell formed by the reduced bases.
+    supercell_fracs = np.dot(supercell.get_positions(), np.linalg.inv(reduced_bases))
     supercell_fracs -= np.rint(supercell_fracs)
+    primitive_fracs = supercell_fracs[list(p2s_map)]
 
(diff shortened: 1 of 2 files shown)


ExpHP:
Also, a note on the speed of sqrt:

Prior to optimizing the function, I instrumented it with IIFEs [as seen here](https://gist.github.com/ExpHP/0fd2e2d3b0d5e74b9d2de9fb8d834255) so I could get more detailed results from the profiler.

The breakdown was:

```
   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
  1192464    1.997    0.000   24.164    0.000 cells.py:501(get_lattice_points)
  1192464   14.287    0.000   20.557    0.000 cells.py:528(filtered)
  1192464    1.244    0.000   13.260    0.000 cells.py:520(do_sums)
  1192464    6.106    0.000   13.040    0.000 cells.py:587(fill_it)
  1192464    3.888    0.000    7.590    0.000 cells.py:516(do_squares)
  1192464    6.267    0.000    6.267    0.000 cells.py:509(do_addition)
  1192464    2.252    0.000    2.252    0.000 cells.py:524(do_roots)  <------
        1    0.000    0.000    0.008    0.008 cells.py:553(prep)
```

Square roots really are surprisingly fast!  Allocations and conditional jumps (e.g. `filtered`) were much worse offenders.  (and of course, the lattice_points array that I should have made a global constant)

atztogo:
Thanks @ExpHP. I could read your comments, but could not have time to reply because I wan in travel. I will review this PR tomorrow.

phonopy/structure/cells.py
```diff
@@ -528,40 +528,75 @@ def _get_smallest_vectors(supercell, primitive, symprec):
...
+
+    # The last final bits are done in C.
+    #
+    # For each list of 27 vectors, we will identify the shortest ones
+    # and move them to the front.
+    shortest_vectors = np.array(shortest_vectors, dtype='double', order='C')
+    multiplicity = np.zeros((size_super, size_prim), dtype='intc', order='C')
+
+    import phonopy._phonopy as phonoc
+    phonoc.gsv_move_smallest_vectors(shortest_vectors,
```
atztogo: If you cam accept, I like to avoid overwriting shortest_vectors if the data size is not very large. This is to let me have attention more when reading this part in the future and this is because I am excellent about forgetting things. I made a branch called ExpHP-gsv-bare-metal and made this update but preserving your comments, https://github.com/atztogo/phonopy/blob/943996cad982397a92370f64d6c27e3aca46d468/c/_phonopy.c#L807 .

ExpHP: Ok.  This is actually the way I was originally going to write it, but then I decided to do it in-place because I thought it would result in fewer cache misses. (however, I didn't measure it; I was only guessing)

I just benchmarked your branch and it is still just as fast, but also cleaner, so this is a good change.  I am about to push an update to this branch with these changes.

PR #52, opened 2017-09-01
Fix bug for FHI-aims wrapper and add handling for multilabels + latex

mospring:


atztogo:
I don't see the purpose of the change in phonopy/phonon/band_structure.py. Could you explain it?

mospring:
The idea is to handle the case of having multiple labels, that means, if I sample a path of non connected points in the brillouin zone, it should still connect them on the plot but put two labels on one "xtick". Does this make things clearer?

On 2 September 2017 00:26:52 CEST, Atsushi Togo <notifications@github.com> wrote:
>I don't see the purpose of the change in
>phonopy/phonon/band_structure.py. Could you explain it?

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity and any typos.

atztogo:
Can you provide a case in a phonopy configuration file for NaCl example for the test?

mospring:
I am not familiar with the configuration file, so I can just guess:

`DIM = 2 2 2
PRIMITIVE_AXIS = 0.0 0.5 0.5  0.5 0.0 0.5  0.5 0.5 0.0
BAND = 0.0 0.0 0.0  0.5 0.0 0.5  0.625 0.25 0.625  0.375 0.375 0.75  0.0 0.0 0.0  0.5 0.5 0.5  0.5 0.25 0.75  0.5 0.0 0.5`

I can't test it because the FHI-aims interface works differently. The path is meant to be:
`GAMMA X U K GAMMA L W X` where U and K are not directly connected but shall be plotted on the same x point.

atztogo:
I don't want to test it by myself. So I will not merge.

Now phonopy supports different force calculators. So it would be easy to make the FHI-aims interface, too. This is the right direction as we discussed.

atztogo:
The change of FHI-aims.py is included in the develop branch. For band_structure.py, it has to be carefully tested because it can affect to many users.

PR #54, opened 2017-09-26
Adapt aims wrapper after change in 0efa91274ce1cf522506a80a9c3129ef7b…

mospring:
…db45c7


atztogo:
Merged to the master branch, but I don't want to maintain phonopy-FHI-aims script. You have to make an FHI-aims interface like the other codes.

PR #56, opened 2017-10-27
RFC: CP2K backend

dev-zero:
This is a Request For Comments PR to get some initial feedback.

Things not tested yet are: everything involving the the 'nac_factor' and the 'distance_to_A' factors.

One also has to install my [cp2k-tools](https://pypi.python.org/pypi/cp2k-tools) package for the CP2K input file parser and output file generator. Inside the phonopy-CP2K-interface I import the CP2K-Tools on the function level to avoid import errors if the CP2K backend is not used.


atztogo:
I realized that cp2k_tools is licensed under the GPL and importing it raises license violation. So I have to remove the line. You can either make cp2k_tools dual license, write an independent CP2KInputParser and dict2cp2k, or ask users to insert the import lines.

dev-zero:
I don't think so, but I relicenced my code under Apache-2.0 anyway

atztogo:
Good! Helpful. There is debate on the treatment of GPL codes. But I choose the more strict one to avoid troubles. As far as I spent some time collecting this related information, more people say that to distribute BSD3 or MIT licensed codes having a line 'import GPL_library' violates GPL license.

dev-zero:
yes, I know. I don't really understand why, but for my cp2k-tools I don't really care, if they are of any use, I'm happy

PR #59, opened 2017-11-27
 Introduction of maximal frequency for ThermalDisplacements class 

JaGeo:
I would like to introduce a maximal frequency for the calculation of the thermal displacements. This would be handy for some analysis of the thermal displacements.  I hope this is okay.

### DIFF
--- a/phonopy/phonon/thermal_displacement.py
+++ b/phonopy/phonon/thermal_displacement.py
@@ -49,9 +49,9 @@ class ThermalMotion(object):
             self._cutoff_frequency = cutoff_frequency
 	  
         if max_frequency is not None:
-	        self._max_frequency = max_frequency
+	    self._max_frequency = max_frequency
         else:
-		self._max_frequency = np.inf
+            self._max_frequency = np.inf
 		
         self._masses = masses * AMU
         self._masses3 = np.array([[m] * 3 for m in masses]).ravel() * AMU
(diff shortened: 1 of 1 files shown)


atztogo:
I will think.

atztogo:
I made max-freq-for-td branch. In it, --fmin and --fmax are used to specify them. The tag names are FMIN and FMAX, respectively. Now --cutoff-freq (CUTOFF_FREQUENCY) is inactive for this calculation. If it works, I will release this.

JaGeo:
Thank you very much for your effort. I have tested it. So far, it works very well. 
Some small suggestions: Maybe you could keep "--cutoff-freq" (CUTOFF_FREQUENCY) in the parser as an alternative to FMIN. It might take some time until everyone has adjusted. Also, FMAX is not printed to the output file so far.

atztogo:
I don't want to put more effort for this since I think this feature is not very popular. So I want users to realize it by themselves reading the document.

JaGeo:
Okay, I see. Keeping "--cutoff-freq" (CUTOFF_FREQUENCY) as an alternative to FMIN still might be a good idea.

atztogo:
I think it's not a good design. It will be a possible source of bugs. I will just show an error message when a user using --cutoff-freq together with thermal displacements.

JaGeo:
Okay. Thanks a lot.

atztogo:
I will write document and then release it. I will close this pull request.

PR #60, opened 2018-01-11
optimize sp2_map generation

ExpHP:
Not much to say here; this is pretty similar to a lot of previous optimization PRs.

### DIFF
--- a/phonopy/structure/cells.py
+++ b/phonopy/structure/cells.py
@@ -290,17 +290,22 @@ class Primitive(Atoms):
             raise ValueError
 
     def _supercell_to_primitive_map(self, pos):
-        inv_F = np.linalg.inv(self._primitive_matrix)
+        frac_pos = np.dot(pos, np.linalg.inv(self._primitive_matrix))
+
+        p2s_positions = frac_pos[self._p2s_map]
         s2p_map = []
-        for i in range(pos.shape[0]):
-            s_pos = np.dot(pos[i], inv_F.T)
-            for j in self._p2s_map:
-                p_pos = np.dot(pos[j], inv_F.T)
-                diff = p_pos - s_pos
-                diff -= np.rint(diff)
-                if (abs(diff) < self._symprec).all():
-                    s2p_map.append(j)
-                    break
+        for s_pos in frac_pos:
+
+            diffs = p2s_positions - s_pos
+            diffs -= np.rint(diffs)
+
+            matches = (abs(diffs) < self._symprec).all(axis=1).tolist()
+
+            assert matches.count(True) == 1
+            p_index = matches.index(True)
+
+            s2p_map.append(p_index)
+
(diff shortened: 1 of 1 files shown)


atztogo:
I'm on travel and don't know how much time to review. I had a quick look. Test fails. To compare with the old one, ``.T`` is needed, 
``frac_pos = np.dot(pos, np.linalg.inv(self._primitive_matrix)).T`` 
but with this fix, test still fails... But this routine should use Cartesian distance, it's my fault.

ExpHP:
Whoops, once again it looks like I forgot to install the package before running unit tests.  Yes, this is very broken.

(I guess it's time for me to learn how to add a `git commit` hook or something...)

phonopy/structure/cells.py
```diff
@@ -290,17 +290,22 @@ def _primitive_cell(self, supercell):
...
-                    break
+        for s_pos in frac_pos:
+
+            diffs = p2s_positions - s_pos
+            diffs -= np.rint(diffs)
+
+            matches = (abs(diffs) < self._symprec).all(axis=1).tolist()
+
+            assert matches.count(True) == 1
+            p_index = matches.index(True)
```
ExpHP: This is wrong.  (it should be using the value of `_p2s_map` at that index)

phonopy/structure/cells.py
```diff
@@ -290,21 +290,25 @@ def _primitive_cell(self, supercell):
...
         for s_pos in frac_pos:
 
-            diffs = p2s_positions - s_pos
-            diffs -= np.rint(diffs)
+            # Compute distances from s_pos to all positions in _p2s_map.
+            frac_diffs = p2s_positions - s_pos
+            frac_diffs -= np.rint(frac_diffs)
 
-            matches = (abs(diffs) < self._symprec).all(axis=1).tolist()
+            cart_diffs = np.dot(frac_diffs, self._primitive_matrix.T)
```
ExpHP: Interestingly, deleting this `.T` does not break any test cases despite being clearly wrong.  This indicates there is a hole in test coverage; I'll try to come up with a test for it.

ExpHP: Update: After looking into this I'm only *more* confounded.  There are already three tests where `_primitive_matrix` is asymmetric, and they have nontrivial-looking positions; yet none of them break when I make this change.

I can't believe that all three of these cases just miraculously succeed by coincidence.

ExpHP: Never mind, this is not too unusual.  In the best case scenario where a difference is exactly zero, multiplying it by *any* matrix will still produce zero.  Multiplying by an incorrect matrix only effects our error estimate for vectors that are "almost zero."

(this while deleting the `.T` is wrong here, it is not surprising that phonopy continues to find exactly one vector with a length near zero, and it would be difficult to write a test case where it did fail)

atztogo:
Thanks. I merged it to develop branch with a minor correction.

PR #61, opened 2018-01-11
optimize POSCAR writing

ExpHP:
I'm a bit surprised to say that POSCAR writing is now the most expensive thing that I see happening when generating displacements.  So here's another quick speedup.

**Note:** Usually I try to preserve the exact behavior of the original code, but in this case I left behind one subtle difference, which is that if a fractional coordinate has a very small but negative value (like `-1e-30`), then it will be written as `1.0000000000000000` instead of `0.0000000000000000`.  This is just how python's modulus works on floats.

I figured it's no big deal, but I mention it just in case. (to get the old behavior, you can add an extra "% 1" to line 280):

```python
# line in the PR; this maps floats into  [0, 1]
unit_positions = scaled_positions % 1

# line which would map floats into  [0, 1)
unit_positions = scaled_positions % 1 % 1
```

### DIFF
--- a/phonopy/interface/vasp.py
+++ b/phonopy/interface/vasp.py
@@ -276,30 +276,32 @@ def get_scaled_positions_lines(scaled_positions):
     return "\n".join(_get_scaled_positions_lines(scaled_positions))
 
 def _get_scaled_positions_lines(scaled_positions):
-    lines = []
-    for i, vec in enumerate(scaled_positions):
-        line_str = ""
-        for x in (vec - np.rint(vec)):
-            if float('%20.16f' % x) < 0.0:
-                line_str += "%20.16f" % (x + 1.0)
-            else:
-                line_str += "%20.16f" % (x)
-        lines.append(line_str)
+    # map into 0 <= x <= 1
+    unit_positions = scaled_positions % 1
 
-    return lines
+    return [
+        " %19.16f %19.16f %19.16f" % tuple(vec)
+        for vec in unit_positions.tolist() # lists are faster for iteration
+    ]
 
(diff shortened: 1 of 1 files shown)


atztogo:
``0.0000000000000000`` is better ``1.0000000000000000`` not to surprise users. This is also expected in a sense of crystallographic convention.

ExpHP:
If you feel strongly about this then I am glad that I brought it up.  I will incorporate the fix.

Just to explain the reasons why I figured it was okay to leave it as is:

* Values like `-1e-30` are pretty hard to come by and easily destroyed during computation.
* It seems to me that any program the POSCAR is given to should be able to handle it correctly, else that program is already broken.
* Reduction of a number into the interval `[0, 1)` is an important concept in many fields beyond just crystallography, yet everybody always writes `x - floor(x)` (equivalent to what I wrote).  Phonopy itself likely contains many, many instances of this already, but this the only place where it's visible (because it is for an output file).
  * There **is** one place where it truly matters, which is code relying on the assumption that `(x % 1.0) // 1.0 == 0`. This assumption is in fact false and I know of one bug in another project which was related to this (it was code that assigned integer `(ix, iy, iz)` indices to images of a site).
  * Most applications really only rely on the fact that `(x // 1.0) + (x % 1.0)` is approximately equal to `x`. This assumption **is** true.

atztogo:
Thanks. I merged to the develop branch. You are logical. For sorting, elements order may not be very important, but I want to keep it.

jochym:
Nice optimizations with switching to more pythonic code ExpHP.

PR #66, opened 2018-05-07
Add remaining character tables to irreps.py

rehnd:
Adds point groups C2, C3v, C6, C3h, D6, T, Th, and O to irreps.py.

Sources used:  https://en.wikipedia.org/wiki/List_of_character_tables_for_chemically_important_3D_point_groups#Dihedral_groups_(Dn)

http://symmetry.jacobs-university.de/

http://www.cryst.ehu.es/cryst/get_point_genpos.html


atztogo:
Thanks. Have you tested those?

rehnd:
I haven't tested them yet, just double checked the values.  Is there a process that has been followed in the past for testing these, or is it just finding an example case for each point group and checking against a reference?

atztogo:
Example cases are expected. Those may be found in http://phonondb.mtl.kyoto-u.ac.jp/. I will do it.

atztogo:
I'm sorry for being late to merge. Finally I merged your pull request. I added tests and fixed some mistakes.

PR #76, opened 2018-07-19
Fhiaims update

Floyd4K:
I updated the FHIaims part of phonopy to be compatible with the PhonopyAtoms class, with python3 and newer matplotlib versions.
The old version was unusable in the last months.

### DIFF
--- a/phonopy/interface/FHIaims.py
+++ b/phonopy/interface/FHIaims.py
@@ -38,6 +38,14 @@
 
 from phonopy.structure.atoms import PhonopyAtoms as Atoms
 
+# FK 2018/07/19
+def lmap(func, lis):
+    """Python2/3 compatibility:
+        replace map(int, list) with lmap(int, list) that always returns a list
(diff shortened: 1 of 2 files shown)


atztogo:
This is not the correct direction.
I said to your colleague that I will not keep scripts/phonopy-FHI-aims in phonopy package. You should simply do the same way as other calculator interfaces to provide FHI-aims users be able to use many features of phonopy.

Floyd4K:
I totally agree. I'm teaching in the way you recommend. However, several people in our group still use the tool and I think it should be at least functional, before it will be removed in the future.

The changes are just in place to make the script functional and keep compatibility. There is no active development.

atztogo:
I see. This time, I will merge them. But I will remove ``script/phonopy-FHI-aims`` as I mentioned to your colleague before at some future.

Floyd4K:
Thank you. I will talk to the respective people and discuss the future of FHIaims integration within phonopy with them.
Florian

Issue #83, opened 2018-10-26
Exclude certain groups of atoms in frozen phonon calc


weixie4:
Dear Professor Togo,

Is it possible that we ask Phonopy to neglect phonon modes for certain groups of atoms in frozen phonon calculations? If so, can you please give me some hint on how to do that? I have this need when studying hybrid inorganic-organic materials, for example, methylammonium lead iodide perovskite (MAPbI3) for which the organic cation MA's phonon modes are not relevant to the material phenomenon under concern. 

I believe many people have similar this need as I do, which is particularly convenient when the irrelevant organic groups contain a large number of C, H, O, N, and S atoms that are not relevant.

So far I can only think of a brutal-force way to cheat Phonopy: 
(1) given a starting structure of MAPbI3 in POSCAR, remove all MA's to obtain an effective structure of PbI3 in POSCAR_eff; 
(2) generate POSCAR_eff-{number} with displaced Pb or I atoms using Phonopy from POSCAR_eff; 
(3) make supercells of POSCAR using the transformation matrix between POSCAR_eff and POSCAR_eff-{number}, add all MA from these supercells back into the corresponding POSCAR_eff-{number}, and obtain POSCAR-{number} files containing all MA and PbI3 atoms;  
(4) calculate the forces of POSCAR-{number} using VASP; 
(5) **remove forces on C, H, N atoms of MA in all disp-{number}/vasprun.xml files to obtain disp_eff-{number}/vasprun.xml;** 
(6) obtain FORCE_SETS from disp_eff-{number}/vasprun.xml using Phonopy and do all the postprocessing as usual. 

It would be nice that there is a simple option in Phonopy that we can specify atoms to neglect so that in steps (2) and (5) Phonopy acts as if these atoms simply don't exist. I understand this feature may not be implemented soon, so I wonder if you could provide me with some suggestions on where to look for in Phonopy to understand how Phonopy parses forces from vasprun.xml.

Thank you and best regards,
Wei

atztogo:
Your detailed description made me understand well what you want.

In this method, forces are parsed from output files of VASP, QE, etc,
https://github.com/atztogo/phonopy/blob/1eab10211c4df3e3ba3d3e2fc603f0bb39b5469b/phonopy/interface/__init__.py#L178

and specifically for VASP,

https://github.com/atztogo/phonopy/blob/1eab10211c4df3e3ba3d3e2fc603f0bb39b5469b/phonopy/interface/vasp.py#L49

[...]

PR #84, opened 2018-11-06
Possibility to build a Hessian for a subsystem in phonopy-FHI-aims

fidanyan:
Support for the FHI-aims '_constrain_relaxation .true._' tag added. Finding this tag after an atom in the 'geometry.in', phonopy-FHI-aims will exclude corresponding atom from the Hessian. It is useful for weakly bonded subsystems to calculate phonons only for one part, but with respect to the presence of another (e.g. vibrations of physically adsorbed molecules).

This feature is highly demanded in our department at the Fritz Haber Institute, and can also be useful for people.

### DIFF
--- a/phonopy/interface/FHIaims.py
+++ b/phonopy/interface/FHIaims.py
@@ -37,7 +37,9 @@
 # POSSIBILITY OF SUCH DAMAGE.
 
 from phonopy.structure.atoms import PhonopyAtoms as Atoms
+import numpy
 
+# fidanyan 2018/11/05
 # FK 2018/07/19
 def lmap(func, lis):
     """Python2/3 compatibility:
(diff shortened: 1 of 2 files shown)


atztogo:
Please read this, https://github.com/atztogo/phonopy/pull/76.

weixie4:
I wish this can be implemented in Phonopy for other calculators e.g., VASP. I had a similar need, as mentioned in #83

atztogo:
First of all, FHI-aims people has to implement the calculator interface. It is the starting point. I told I will not support phonopy-FHI-aims script any more.

On this feature, as I wrote in #83, this has to be done carefully. First, this has to work robustly, otherwise we will receive many questions on this when it failed. So this is not just an addition of a simple feature, but it's almost a research or development of method.

atztogo:
@fidanyan, in the develop branch (I think it will be soon released), I worked to make making a calculator interface easier. Most of the modification you need is collected in `phonopy.interface.__init__.py`. You have already parsers and writers, so the job will not be tough.

fidanyan:
@atztogo, thank you for the information! I will discuss with @flokno and colleagues how it better should be done.

atztogo:
One more file you have to modify is `phonopy/cui/phonopy_argparse.py`, where the modification is obvious.

PR #92, opened 2019-03-26
Improved TURBOMOLE interface

anttikarttunen:
Dear Togo,

Sorry for breaking the previous build. I was only running test_TURBOMOLE.py in the test/interface directory, which was not enough. 

I realized that referencing external files in the unit tests is not straightforward. So, I made the TURBOMOLE interface bit more flexible and modified the test. Now "python setup.py test" ran all tests successfully.

Best wishes,
Antti


atztogo:
Merged to develop branch! Thanks.

PR #96, opened 2019-06-19
_set_heat_capacity_P_polyfit: missing factor 2 in dvdt

ezanardi:
method QHA._set_heat_capacity_P_polyfit : 
if equilibrium volumes vs T is fitted to a polynomial of degree 2, then dvdt (partial derivative of the equilibrium volumes with respect to the temperature) should be parameters[0] * 2 * t + parameters[1]


atztogo:
Thank you very much. Merged to develop branch. If I remember correctly, this bug didn't exist before.  So I introduced it at some moment. Anyway this is a critical bug. I hope people use `Cp-temperature.dat` but not `Cp-temperature_polyfit.dat`.

PR #97, opened 2019-07-26
add pyproject.toml to declare numpy as build dependency

flokno:
Solves https://github.com/atztogo/phonopy/issues/53

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,2 @@
+[build-system]
+requires = ["setuptools", "numpy"]


atztogo:
Do you know what are these warnings appearing in pip install?

```
  WARNING: Missing build requirements in pyproject.toml for file:///Users/togo/code/phonopy.
  WARNING: The project does not specify a build backend, and pip cannot fall back to setuptools without 'wheel'.
```

flokno:
You are right, I forgot to include `wheel` in the requirements, I'll add them.

flokno:
I replace this PR by https://github.com/atztogo/phonopy/pull/98

atztogo:
I just don't know these stuffs. Thanks.

flokno:
You're welcome! I didn't know about this as well, until recently when I started using `poetry` to package my python projects. `poetry` is a great tool, but also very strict and picky with dependencies, which forced me to look into the business of python packaging a bit more.
But also I'm far from being an expert, for sure.

PR #100, opened 2019-10-10
CP2K backend: updates and documentation

dev-zero:
* switch to my new `cp2k-input-tools` module as parser backend for the CP2K input file for full syntax support
* implement alternative cell specifications (ABC/ALPHA_BETA_GAMMA or A/B/C)
* properly use CP2K's base units instead of mangling them
* use the `iter_collect_forces` to always pick up the last ones in the `.xyz` files in case the user ran the calculation multiple times
* add tests and documentation
* add NaCl-CP2K example


atztogo:
Merged to develop. Thanks.

PR #104, opened 2019-11-15
spelling error when outputting FC to phonopy.yaml

pjcon:
When writing force constants file to phonopy.yaml, raises AttributeError. Fixed a spelling error by adding an underscore in the function name call and now it works.


atztogo:
Thanks. Merged.

PR #106, opened 2020-01-23
small changes to hiphive interface

erikfransson:
small changes to hiphive interface

### DIFF
--- a/phonopy/interface/hiphive_interface.py
+++ b/phonopy/interface/hiphive_interface.py
@@ -37,12 +37,23 @@ from ase.atoms import Atoms
 from hiphive import ClusterSpace, StructureContainer, ForceConstantPotential
 from hiphive.fitting import Optimizer
 from hiphive.cutoffs import estimate_maximum_cutoff
+from hiphive.input_output.logging_tools import set_config
+set_config(level=30)
+
+
+def phonopy_atoms_to_ase(atoms_phonopy):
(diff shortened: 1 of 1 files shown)


atztogo:
Merged. Thanks.

erikfransson:
@atztogo I tested the `get_fc2` function but never by letting phonopy call it. I think maybe this pull request broke it.

I added that `get_fc2` (in `hiphive_interface.py`) takes `symprec` as input argument but I forgot that the phonopy code that calls `get_fc2` thus needs to pass along `symprec`.

I dont want to touch the coding regarding the other fc calculators but in `get_fc2` in `fc_calculator.py` one could maybe allow  **kwargs for additional arguments that each calculator may need from phonopy?

atztogo:
@erikfransson , sorry for my slow reply.
It's fine. Could you proceed it with ``**kwargs`` as you like if you are OK? You should be better than me in coding.

Please make a branch from develop. The docstring of ``get_fc2`` in ``fc_calculator.py`` will be moved to that in ``alm.py``. This ``get_fc2`` is called only at https://github.com/atztogo/phonopy/blob/a359d682b3084bbd6f5b94b7e971263079d7e063/phonopy/api_phonopy.py#L2806 if I remember correctluy. I used this ``get_fc2`` in phono3py, too, but you don't need to bother from it.

PR #108, opened 2020-01-29
Ft add summary file output data flags

pjcon:
Update Phonopy user interface flags/settings to output extra data to phonopy.yaml, including: force constants, force sets, dielectric tensor, born effective charges, displacements.

Flags (store_true)/settings (set to .TRUE.): --include-fc/INCLUDE_FC, --include-fs/INCLUDE_FS, --include-eps/INCLUDE_EPS, --include-bec/INCLUDE_BEC, --include-disp/INCLUDE_DISP, --include-all/INCLUDE_ALL.

I have also fixed a few parts that caused bugs in `class PhonopyConfParser`, and slightly changed how output options are sent through `finalize_phonopy` to `PhonopyYaml`, but the change is relatively transparent.


pjcon:
I have included a small stub test called `test_write_phonopy_yaml_extra` in interface/test_phonopy_yaml.py but I'm not sure how rigorous you want it in comparison to the current test.

atztogo:
Thanks for this nice option and  also for fixing my mistakes. One thing, I want to discuss. You separate Born effective charges and dielectric constants, but in phonopy they are used as a couple always, so maybe one option for both is convenient. How do you think?

pjcon:
Ok. I'll merge them at the front end using the flag `--include-born`, but I'll keep them separate on the back end to maintain individual get/set control for each output data type in cui/settings.py

atztogo:
Thanks. Merged to develop branch.

atztogo:
I close this PR.

PR #109, opened 2020-02-08
Improve hiphive interface

erikfransson:
I've made hiphive work through the CUI, including passing along symprec (--tolerance) to hiphive.

I added an example, mostly for my own sake to check that everything works as expected, which I can remove if you like.


atztogo:
Thanks. Merged to develop.
ALM also accepts symmetry tolerance (internally using spglib), but I don't yet write the python wrapper to set it. I should do it.

PR #113, opened 2020-02-27
TURBOMOLE example: renamed pre-calculated data directory

anttikarttunen:
TURBOMOLE data directory supercell-001 renamed to example-001 to avoid overwriting pre-calculated data while running the example.


atztogo:
Thanks!

PR #117, opened 2020-05-04
FIX: bandplot | fix when `output_filename` is given

flokno:
`phonopy-bandplot band.yaml -o band.pdf` produces an error. This PR fixes that error.


atztogo:
Thanks!

PR #122, opened 2020-06-25
Spin polarization support and test for CASTEP interface added

ladyteam:
Spin polarization support added to CASTEP interface. Test for CASTEP interface is implemented. The only thing is left is examples.


atztogo:
Thanks. Merged.

PR #125, opened 2020-07-27
Update phonopy-bandplot :: multiple arguments like --xlabel, --ylabel, --title, --points, --line are working now.

kgmat:
Updated phonopy-bandplot(mostly re-adjustment of your original one), for a number of arguments that were not working earlier.

arguments now working:

--xlabel, --ylabel, --title, --line, --points(also added points type) 

Thank you Prof. Togo for this wonderful code Phonopy with amazing user-interface.

Regards

### DIFF
--- a/scripts/phonopy-bandplot
+++ b/scripts/phonopy-bandplot
@@ -1,5 +1,6 @@
 #!/usr/bin/env python
 
+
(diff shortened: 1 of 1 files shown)


atztogo:
Nice. Merged. Thanks! Frequency unit can be chosen by users. I will remove the unit later.

PR #127, opened 2020-09-09
Include structural and band/mesh information in gruneisen.yaml

ab5424:
We aim at different pymatgen classes to plot a Grüneisen parameter band structure and/or mesh. To evaluate the data, we need the structure. Therefore, it would be convenient to have it present in the gruneisen.yaml file. 

The majority of the code in this pull request was taken from the phonon band structure and mesh classes. The resulting gruneisen.yaml is similar to mesh.yaml for a gruneisen mesh and similar to band.yaml for a gruneisen band structure. It also adds compression for the gruneisen.yaml files.

I can try to include a test if necessary.

### DIFF
--- a/phonopy/gruneisen/band_structure.py
+++ b/phonopy/gruneisen/band_structure.py
@@ -32,6 +32,9 @@
 # ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 # POSSIBILITY OF SUCH DAMAGE.
 
+import sys
+import gzip
+import yaml
 import numpy as np
 from .core import GruneisenBase
 from phonopy.units import VaspToTHz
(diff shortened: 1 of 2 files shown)


atztogo:
Merged. Thanks for your careful updates.

I haven't touch gruneisen for a while. Recently I got a question on the treatment of divergence near Gamma for acoustic modes (probably solvable by analytical expression), but I have no time to work on gruneisen currently. Anyway, tests are needed to touch the code from time to time.

So I'm sorry that no tests were written for gruneisen. It is helpful if you could write the test, in case if you have some time. The tests are in `phonopy/test` directory. Just now I added `conftest.py` so you can write tests in pytest if you like. You can copy the input data in `example/*-gruneisen` in `phonopy/test` as you want.

JaGeo:
We plan to further work with and maybe also on the gruneisen script. @ab5424 and I are actually working on this together.

PR #134, opened 2020-11-10
fix for  "--gnuplot should not require --legacy"

kgmat:
Dear Dr. Togo,

  This PR is about fixing of "--gnulot should not require --legacy"

  Thanks & Regards

### DIFF
--- a/scripts/phonopy-bandplot
+++ b/scripts/phonopy-bandplot
@@ -478,6 +478,33 @@ def plot(args):
             filenames = args.filenames
         bands_data = [read_band_yaml(fname) for fname in filenames]
 
+    if args.is_gnuplot:
+
+        distances = bands_data[0][0]
+        frequencies = bands_data[0][1]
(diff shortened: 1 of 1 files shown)


atztogo:
Thanks, @kgmat. Now there are two gnuplot output routines, one in `def plot` and the other in `def old_plot`. Do you think they can be merged to one?

kgmat:
@atztogo, I was actually thinking of that but ended with keeping them in both the functions. I will try to keep it in one place. Thanks

atztogo:
@kgmat , thanks for your consideration. I understand `phonopy-bandplot` is so messy, fragile, and there is no test. It's my fault. So please modify as you like. You can break it and of course, it is no need to be perfect at all.

kgmat:
Dear Dr. Togo, as you suggested. Now a separate function is there in `phonopy-bandplot` for writing gnuplot-style output. 

Thanks

atztogo:
Great. Merged. Thanks.

PR #150, opened 2021-05-31
Added "phonopy-totaldosplot" to plot multiple total-DOS together, with line properties

kgmat:
Dear @atztogo,

     A user recently asked about plotting multiple total-DOS together in the mailing list.

     Since a similar feature is already there for bandstructure in phonopy, having the same for DOS would be good.

    This PR adds one script "phonopy-totaldosplot" which can handle multiple totalDOS files together, and set line properties(width, color, style) for each file. 

    Regards

### DIFF
--- a/scripts/phonopy-totaldosplot
+++ b/scripts/phonopy-totaldosplot
@@ -1,0 +1,272 @@
+#!/usr/bin/env python
+
+# Copyright (C) 2011 Atsushi Togo
+# All rights reserved.
+#
+# This file is part of phonopy.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions
+# are met:
+#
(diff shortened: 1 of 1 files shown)


atztogo:
Thanks @kgmat, but, total DOS can be plotted using phonopy-pdosplot. If I include this PR, we have to support this. To avoid increasing supporting effort, it is rather better to have this feature in phonopy-pdosplot.

In addition, on PR, pylint supports Coding Standard check (https://www.pylint.org/), though I use flake8, so could you have a look at the coding standard of python? No need to be very strict, but if you could follow basic ones like no speces next to parenthesis, spaces after ",", 80 characters in one line, etc, it is helpful.

kgmat:
@atztogo, I will try to include it in phonopy-pdosplot.

And thanks for suggesting coding standards, will keep up with it in the future PRs.

Regards

PR #159, opened 2021-10-16
Small changes for Fleur interface

neukirchen-212:
I had some stuff lying around and since I need to define a variable as mentioned by @atztogo, it is a good point to merge again.

Changes:
- I was a bit overzealous in saying Fleur can do DFPT phonons. We are still on that.
- N defined properly at @atztogo 's request
- 'factor' and 'shift' features from the Fleur input generator can now be used
- 0.0 scales in the inpgen can be properly used 1.0s

Best, Alexander Neukirchen


atztogo:
Thanks, @neukirchen-212. I think `N`=1 in your change of `calculator.py` (`additional_info` does not exist in the scope), and I did so. I merged your PR to the refactoring branch currently I am working on, which will be merged to the develop branch.

PR #193, opened 2022-04-01
Fix for fetching VASP version correctly.

kgmat:
For VASP version, one case had a line in vasprun.xml like: **5.4.4.18Apr17-6-g9f103f2a35**

This line had "-" as well as non integers, which clashes with [this line](https://github.com/phonopy/phonopy/blob/5953d7963dbdd7abdf20e5e6904f5098f6a49c40/phonopy/interface/vasp.py#L580).

and throws and error, **ValueError: invalid literal for int() with base 10: '18Apr17-6-g9f103f2a35'**

This PR fixes the issue above.

Thanks


atztogo:
Thanks @kgmat. Merged.

PR #203, opened 2022-06-06
Fix formulation of eigenvectors

Cloudiiink:
Add the missing imaginary unit in the section 'Create modulated structure'


atztogo:
True. Thanks @Cloudiiink for your PR!

PR #265, opened 2023-03-06
NAC by Wang in API support

ladyteam:
Hi Togo!
Phonon calculations with NAC using method by Wang et al. was implemented in PHONOPY a while ago. We can used by calling phonopy script and an option  --nac-method. I was found a problem trying to make calculations with Wang method using PHONOPY API. To fix the issue, I've made a code modification by adding nac_method parameter of the class Phonopy. Please consider this PR to be accepted in development branch. Thank you!

### DIFF
--- a/phonopy/api_phonopy.py
+++ b/phonopy/api_phonopy.py
@@ -142,6 +142,7 @@ class Phonopy:
         supercell_matrix=None,
         primitive_matrix=None,
         nac_params=None,
+        nac_method=None,
         factor=VaspToTHz,
         frequency_scale_factor=None,
         dynamical_matrix_decimals=None,
@@ -205,6 +206,7 @@ class Phonopy:
                 DeprecationWarning,
             )
         self._nac_params = nac_params
+        self._nac_method = nac_method
         self._dynamical_matrix_decimals = dynamical_matrix_decimals
 
         # set_band_structure
@@ -3646,6 +3648,8 @@ class Phonopy:
             )
             nac_params = self._nac_params.copy()
             nac_params.update({"born": borns, "dielectric": epsilon})
+            if self._nac_method is not None:
+                nac_params.update({'method': self._nac_method})
         else:
             nac_params = self._nac_params
 
--- a/phonopy/api_phonopy.py
+++ b/phonopy/api_phonopy.py
@@ -3649,7 +3649,7 @@ class Phonopy:
             nac_params = self._nac_params.copy()
             nac_params.update({"born": borns, "dielectric": epsilon})
             if self._nac_method is not None:
-                nac_params.update({'method': self._nac_method})
+                nac_params.update({"method": self._nac_method})
         else:
             nac_params = self._nac_params
 
--- a/phonopy/cui/load.py
+++ b/phonopy/cui/load.py
@@ -51,6 +51,7 @@ def load(
     unitcell=None,
     supercell=None,
     nac_params=None,
+    nac_method=None,
     unitcell_filename=None,
     supercell_filename=None,
     born_filename=None,
@@ -228,6 +229,7 @@ def load(
         )
         _calculator = calculator
         _nac_params = nac_params
+        _nac_method = nac_method
         _dataset = None
         _fc = None
     elif phonopy_yaml is not None:
@@ -277,6 +279,7 @@ def load(
         store_dense_svecs=store_dense_svecs,
         use_SNF_supercell=use_SNF_supercell,
         calculator=_calculator,
+        nac_method=_nac_method,
         log_level=log_level,
     )
 


atztogo:
The method choice can be written in `nac_params` dictionary. Each method may require different data input, so use of dictionary with the method identifier is reasonable. How about using this?

https://github.com/phonopy/phonopy/blob/dac918000714d697ac48cf7deb2ec546d4384d8a/test/harmonic/test_dynamical_matrix.py#L533

The problem is that the documentation is not well written about this.

atztogo:
You can create `Phonopy` instance without setting `nac_params`. Then you can set `nac_params`. So you don't need to recalculate dynamical matrix. If it is not so implmented, it is a bug.

`nac_params` is defined by a dictionary, because the parameters required for NAC may be different for different method. So I think `method` should be coupled with parameters.

ladyteam:
Dear Togo,
Please correct me if I'm wrong. But after reading the code and come testing I have found, that `phonopy.load` function calls (twice for some reason, but this is separate question) `_set_dynamical_matrix` function of `Phonopy` class and returns `Phonopy` instance wich contains _dynamical_matrix. The last is the object constructed by `get_dynamical_matrix` function wich imported from `phonopy.harmonic.dynamical_matrix`. The function `get_dynamical_matrix` returns either `DynamicalMatrix` or `DynamicalMatrixGL` or `DynamicalMatrixWang` depending on values in `nac_params` dictonary. 

All said above means that changing nac parameters after calling `phonopy.load` do nothing untill `ph._set_dynamical_matrix()` call.

An examples:
`ph = phonopy.load(is_nac=True)` 
will return an instance of `DynamicalMatrixGL` in `ph._dynamical_matrix`

`ph = phonopy.load(is_nac=True)` 
will return an instance of `DynamicalMatrix` in `ph._dynamical_matrix`
`ph = phonopy.load(is_nac=True)` 
`ph.nac_params["method"]="wang"`
will return an instance of `DynamicalMatrixGL` in `ph._dynamical_matrix`

And following 3 lines of code:
`ph = phonopy.load(is_nac=True)` 
`ph.nac_params["method"]="wang"`
`ph._set_dynamical_matrix()`
will return an instance of `DynamicalMatrixWang` in `ph._dynamical_matrix` as required.

I am agree this is not a real calculations but assignements of Force Constants Born and NAC parameters, so it is not time consuming, but memory demanding.

I can provide (somehow) tests to show that in case of Wang one need to call `ph._set_dynamical_matrix()` after changes in nac_params dictonary to switch to Wang NAC threatment mechanism if you wish.

Thank you in advance.

atztogo:
Wang's method can be considered as an approximation of the Genze&Lee method. In general, the latest is better physically. So the later is the default. I even consider to drop support of the former method.

`phonopy.load` mainly  aims to read `phonopy_*.yaml` as input. If you want to specify more detailed setting, it is recommended to use `Phonopy` class. 
I also expect `phonopy.load` as a short cut to create `Phonopy` class instance. In this case, it is designed to use some default settings intentionally.

`Phonopy.nac_params = nac_params_dict` invokes the setter method:
https://github.com/phonopy/phonopy/blob/dac918000714d697ac48cf7deb2ec546d4384d8a/phonopy/api_phonopy.py#L747-L751
where `self._set_dynamical_matrix()` is called. You call `Phonopy.nac_params['method']`. In this case, getter method is called. This is not expected usage.

I assume computational overhead of `get_dynamical_matrix` is small.
https://github.com/phonopy/phonopy/blob/dac918000714d697ac48cf7deb2ec546d4384d8a/phonopy/harmonic/dynamical_matrix.py#L1115
It is not a good idea to call several times if it is not the case. Please indicate it in the different github issue.

ladyteam:
Surprisingly, Wang's method appears to be more accurate than default Gonze&Lee one in case finite difference supercell approach. LO-TO splitting have a physical sense in proximity of Gamma point. Therefore in other points than close to Gamma the dispertion of phonon branches calculated with NAC term should coincide with phonon branches calculated without NAC, and descripancy is significant in case of NAC method by Gonze&Lee. Therefore Wang is more accurate in practice.

`phonopy.load` is excelent to work with raw data (FORCE_CONSTANTS/FORCE_SETS and POSCAR, BORN). It is very intelligent way to write external tools. Selecting the method for NAC with `phonopy.load` seems to be very logical and indeed wellcome for users (I suggest).

Ok now I see that:
>  Phonopy.nac_params = nac_params_dict invokes the setter method:

So now there are two ways to change default NAC method:
1. not perfect but works
`ph = phonopy.load(is_nac=True)`
`ph.nac_params["method"]="wang"`
`ph._set_dynamical_matrix()`

And
2.
`ph = phonopy.load(is_nac=True)`
`nac_params=ph.nac_params`
`nac_params["method"]="wang"`
`ph.nac_params=nac_params`

First code is 3 lines, second is 4. Isn't it more elegant to have zero additional lines instead?
`ph=phonopy.load(is_nac=True, nac_method='wang')`

There is nothing special in setting NAC method indeed it is usefull to have an option. I suggest, that current implementation of NAC method selection in API is a bit tricky and may rise users questions in future. In any case, thanks for details explanation.

atztogo:
The method is coupled with other parameters (dielectric constants, etc), and the kinds of parameters also depend on the method. So I don't think it is a good idea to have `nac_method` parameter.

```
ph = phonopy.load(is_nac=True) # is_nac=True is default.
```
is expected to read `BORN` file. Or unless `BORN` exists, NAC parameters are looked for in `phonopy_*.yaml`. No NAC information is found, even with `is_nac=True`, NAC is not invoked.

I want to squeeze those information in a data structure, which is easier for me to maintain the phonopy code. For command user interface and `phonopy.load`, the method choice is better integrated in the format of `BORN` or `phonopy_*.yaml`. YAML is a structure format, so integrating `nac_method` in `phonopy_*.yaml` is easy. Although design of current `BORN` format is bad, we may be also also possible to support `nac_method` in `BORN` file.

ladyteam:
To be honest, I hate yaml as a format for storing data and parameters. First of all, it is text format and therefore unsafe. One may occasionally edit yaml text file and bring errors or even destroy the data. There are huge amounts of white spaces and the size of a file can be sometimes huge.  Hdf5 container is far much better and modern and more compact. It's only my personal opinion, of course. And I wouldn't recommend to touch BORN file for many reasons, it's perfectly formatted.

The `phonopy.load` is an outstanding tool to work with `PHONOPY` (except the choice of NAC method ^__^). It is compact and simple and powerful! I had never any needs to work with phonopy_*.yaml, the less additional files used the fewer bugs in development, and the easier life for users. 

Actually, it was not my idea to rise Wang NAC method. I was happy with Gonze&Lee before. It was my colleagues who asked me to resolve the issue with nac_metod in `PHONOPY` API. I did it and they are happy. Initially, I wanted to bring nac_method with nac_parameters dictionary, but it seems impossible to do without significant efforts. That's why I added the new parameter. Now I see it is a kind of philosophical question, and hence we'll gonna use an approach as you kindly suggested (setter method), no problem. Thanks for explanations. I close this PR until next demand on NAC method choice with `phonopy.load` (if any). Still, the discussion was useful (at least for me).

atztogo:
For `BORN`, for example, we can put a trigger at the first line to support different format. The implementation is easy for me.

ladyteam:
I think modification of `BORN` file can rise questions from users. You see, some of them can't even put options in command line properly... 

The simplest way is to modify comments in the following manner:
Optional. Means to provide NAC parameters and their priority:                                                                       
        1. born_filename
        2. nac_params *
        3.  3. phonopy_yaml.nac_params if existed and is_nac=True.
        4. 'BORN' is searched in current directory when is_nac=True.
        * Nac method by Wang can be triggered as shown in test_dynmat_wang(ph_nacl) (see phonopy/test/harmonic/test_dynamical_matrix.py)
        
Really, I don't know. It would be better to discuss with enhanced users.

atztogo:
`--nac-method` option is finally not a nice idea.  We may have more different methods for NAC in the future. I never thought that phonon calculation of this type would become so popular like currently, and we would be interested in more detailed methods at that time.

`--nac-method` option can be deprecated. So I think if it is set in `BORN`, it should be preferred when used from `phonopy` script. How do you think? Once we agree, I can code it.

ladyteam:
Indeed, LO modes become very popular these days, probably due to complicate nature of this type of vibrational modes. After all, I think `BORN` file is a good place to store the type of NAC method. Moreover, since BORN file already contains dielectric linear properties (first order dielectric susceptibility (epsilon) + BORN charges) it can potentially store nonlinear susceptibility tensor in the future, which is used in spectroscopy. I understand spectroscopy is external tool but who knows... So maybe it is worth to be more radical and replace `BORN` by `born.yaml`, i.e. dielectric properties will be stored in `yaml` language. Why not? `BORN` is already a text file and `yaml` is human-readable and can be easily extended in functionality in future (if needed) to store new data with dielectric properties. It's just a suggestion, of course.

atztogo:
OK. I will make an experimental implementation of `born.yaml`.

atztogo:
@ladyteam, please have a look at PR #269.

Issue #263, opened 2023-03-01
NAC not working with vasprun.xml


mystery-e204:
Hi,

I tried running phonopy to plot the phonon dispersion of a finished VASP run but when I try setting `NAC = .TRUE.` it does not work. I followed the [VASP-DFPT tutorial](https://phonopy.github.io/phonopy/vasp-dfpt.html) but I used finite differences instead of perturbation theory.

Here is what I did:
1) Run VASP in a supercell (4x4x4) with `IBRION = 6` to generate the Hessian matrix in vasprun.xml.
2) Run `phonopy -fc vasprun.xml` to generate FORCE_CONSTANTS file.
3) Provide
   - POSCAR-unitcell
   - BORN
   - band.conf
4) Run `phonopy -c POSCAR-unitcell --dim="4 4 4" --readfc -p band.conf`

This is in my band.conf:
```
BAND = ...
BAND_POINTS = 61
FC_SYMMETRY = .TRUE.
PRIMITIVE_AXES = AUTO
NAC = .TRUE.
```

This is the result:
<img width="285" alt="image" src="https://user-images.githubusercontent.com/43782561/222258090-3b7e2b9e-8452-464e-ae30-193d28e3de60.png">
There is no LO-TO splitting due to NAC. I might be doing something wrong but I never got any errors and the output says:
```
Force constants are read from "FORCE_CONSTANTS".
...
NAC parameters were read from "BORN".
```

atztogo:
Using [this example](https://github.com/phonopy/phonopy/tree/develop/example/NaCl) and `band.conf`:
```
DIM = 2 2 2
PRIMITIVE_AXIS = AUTO
BAND = 0.0 0.0 0.0  0.5 0.0 0.0  0.5 0.5 0.0  0.0 0.0 0.0  0.5 0.5 0.5
BAND_POINTS = 61
FC_SYMMETRY = .TRUE.
NAC = .TRUE.
```

I ran phonopy (develop branch 5d9f37e9, and also latest conda package macOS ARM64)
```
% phonopy -c POSCAR-unitcell band.conf -p
        _
  _ __ | |__   ___  _ __   ___   _ __  _   _
 | '_ \| '_ \ / _ \| '_ \ / _ \ | '_ \| | | |
 | |_) | | | | (_) | | | | (_) || |_) | |_| |
 | .__/|_| |_|\___/|_| |_|\___(_) .__/ \__, |
 |_|                            |_|    |___/
                                      2.17.1

[...]

PR #266, opened 2023-03-09
Improve parsing of Hessian units in vasprun.xml

mystery-e204:
Based on the discussion in #263, I have created a pull request that should improve the parsing of the `vasprun.xml` file for newer versions of VASP. Instead of using the version number, I use an additional XML tag on the `<dynmat>` node with the name `unit` to determine the units of the Hessian matrix in the XML. This should be more robust.

In addition, I have changed two things in the code:

1) Improve the clarity of the code by using XPath Syntax
2) Allow for an early exit from parsing after the information has been acquired

I hope that this is ok. Otherwise I can provide a more specialized merge request that only changes the absolute minimum required.

### DIFF
--- a/phonopy/interface/vasp.py
+++ b/phonopy/interface/vasp.py
@@ -585,16 +585,9 @@ class Vasprun:
 
         """
         fc_tmp = None
-        version_nums = None
+        hessian_units = ""
         num_atom = 0
         for event, element in vasprun_etree:
-            # VASP version
-            if element.tag == "generator":
-                for element_i in element.findall("./i"):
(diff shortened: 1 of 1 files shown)


atztogo:
At some versions of VASP6, hessian unit may not be written in `vasprun.xml`, e.g., VASP-6.1.0. Does this PR work for it?

mystery-e204:
Yes, I checked the git log on the file where the Hessian and the units are written. The change to THz coincides with the new `unit` tag, so it should always work this way.

atztogo:
Thanks!

PR #276, opened 2023-05-15
Refactor: abacus interface

1041176461:
1. Fix: magnetic settings in STRU
2. Refactor: delete setting.conf needed
3. Feature: add ABFs supported
4. Feature: add exception handling


atztogo:
@1041176461, I see your commit doesn't affect to those outside of the abacus interface, so I can merge it. But I recommend you to write tests. Since now you know what was the issue and what is the change, it is easier to write tests, and as you know tests will help you a lot in the future. Do you want me to merge it right now?

1041176461:
> @1041176461, I see your commit doesn't affect to those outside of the abacus interface, so I can merge it. But I recommend you to write tests. Since now you know what was the issue and what is the change, it is easier to write tests, and as you know tests will help you a lot in the future. Do you want me to merge it right now?

Thanks for your advice. I will write tests for the interface in this PR, recently. So it can be merged after I finish writing and I will assign to you.

1041176461:
I found I have written "/test/interface/test_abacus.py" when I submitted my first PR. I modified it to match this new PR. I'm wondering if this simple test function is enough?

atztogo:
I think it is good.

FYI, you may want to compare cells in writing tests in the future. To compare two cells, you can use `isclose` function for a simple comparison (it seems your test could be written using this),

https://github.com/phonopy/phonopy/blob/210f2455d8e0d3d0e29d50aedb2d0dc00dcd172c/phonopy/structure/cells.py#L896-L903

or a bit complicated case, a pytest fixture

https://github.com/phonopy/phonopy/blob/210f2455d8e0d3d0e29d50aedb2d0dc00dcd172c/test/conftest.py#L260-L261

atztogo:
Merged. Thanks.

PR #284, opened 2023-06-01
[WIP] Extension of the Grüneisen scripts

JaGeo:
Dear @atztogo , 

I have started to work a bit on the Grüneisen scripts. I would like to update the format of the yaml files (similar to the normal band structure outputs), add the option to use automatic k-path determination, add options to add eigenvectors, enable band connection options etc. It's still in a very early stage and very hacky. We currently struggle to get output files with correct band connection. To correct it manually, we would also need the eigenvectors in the output files. (We could, of course, do everything via the API.)

What do you think about this? Would there be interest from your side? Are you planning to do something similar? Do you have any suggestions?

Best, 
JG


atztogo:
@JaGeo , thanks for your contribution. 

Band connection is not as easy, but this is probably another issue to discuss.

> What do you think about this?

Good.

> Would there be interest from your side?

I haven't used it for long time.

> Are you planning to do something similar?

No plan.

> Do you have any suggestions?

- We should write more tests.
- Development (including maintenance) of user interface (i.e., `phonopy-gruneisen`) is tough. Only supporting `phonopy_params.yaml` type file as input is a possible choice to ease it. To read the yaml type phonopy file, `phonopy.load` can be used.
- Writing documentation along with some examples is important.

JaGeo:
Thank you so much for the suggestions.
- I think I will start by changing the format of the "gruneisen.yaml" files (so that it agrees with the band.yaml file format) and add tests. Even completely breaking the current format did not lead to any test errors so that it is clear that more tests are needed here.
- The next step would be trying to change the scripts as well. 

This may take a while.

JaGeo:
I have updated the list of todos.

scripts/phonopy-gruneisen
```diff
@@ -333,9 +334,22 @@ def main(args):
 
             mpl.use("Agg")
 
-    if args.band_paths:
-        from phonopy.phonon.band_structure import get_band_qpoints
-
+<<<<<<< HEAD
```
atztogo, line 337: Could you fix this?

atztogo:
During you have still the code in your mind, it would be helpful if you could write tests on the Gruneisen code?

atztogo:
Codecov comments may be good to read, https://github.com/phonopy/phonopy/pull/284/files, but since originally not much tests were written, so you don't need to write all the tests against suggested by codecov.

You added lines `#TODO: ...`. If you want to leave comments, please write them in the docstring in a readable way to other people. If it is only for you, please remove them. The information that you want to tell can be also mentioned in this PR or issues on github.

atztogo:
Probably some people may rely on the old format. Could you write how the output format is changed in the documentation (https://phonopy.github.io/phonopy/gruneisen.html#phonopy-gruneisen)?

atztogo:
There are two examples:
* https://github.com/phonopy/phonopy/tree/develop/example/NaCl-gruneisen
* https://github.com/phonopy/phonopy/tree/develop/example/Si-gruneisen
Please have a look at them and update them if needed.

JaGeo:
Thank you for the comments. This is very helpful! I will keep all of this in mind. It will take a few weeks until all of this is finished. 🙈

The TODOs are only for me. I will remove them before your code review!

atztogo:
@JaGeo, please try to use `pre-commit`. After installation of pre-commit package, on the phonopy home directory, you can run
```
pre-commit install
```
Then, when you `git-commit`, pre-commit hook runs automatically, by which the code is formatted and python module imports are reordered, etc. Without using pre-commit locally, as you can see [pre-commit.ci] service runs to do the job, but in this case, you have to merge the commit by [pre-commit.ci] service to your local branch.

JaGeo:
I fear I need to close it for now. I don't have the capacity to finish it at the moment. Really sorry.

PR #309, opened 2023-10-19
Fix:  cancel the default setting of magnetic moment for abacus interface

1041176461:
1. cancel the default setting of magnetic moment for abacus interface
2. modify name of supercell file from 'sSTRU' to 'STRU.in'


1041176461:
@atztogo Hello, could you please tell me how to fix four errors reported by Codecov? Thanks!

1041176461:
Thanks! I have added the test function.

atztogo:
Merged. @1041176461, thanks.

You might not run pre-commit before commit. For example in phonopy, https://github.com/phonopy/phonopy#pre-commit
I recommend running it locally because it is useful.

PR #337, opened 2024-03-11
Fix calculation of q_direction for band struture with NAC

chazeon:
This pull request enables the NAC for the entire phonon band path. 
Without this fix, the NAC is applied to the $\Gamma$ point only.

### Details

Before commit c2b1002, the `q_direction` is passed in as an argument of the `_solve_dm_on_path` function; after commit c2b1002 because no `q_direction` can be supplied, NAC is applied to $\Gamma$ point only.


chazeon:
Without this fix:
image
after the fix:
image

PR #356, opened 2024-05-08
2D dipole-dipole corrections

bastonero:
Here I implement a way to obtain the phonon dispersion of 2D materials using the correct screening for the non-analytical correction. In fact, the formula differs from the 3D bulk, as demonstrated in the pioneering paper: T. Sohier _et al_., _Nano Lett._ 2017, 17, 6, 3758–3763 (DOI: 10.1021/acs.nanolett.7b01090). A flexible code is provided to support 1D screening (to be implemented) as recently derived in Rivano, N. _et al._,  _npj Comput. Mater._ 9, 194 (2023) (DOI: 10.1038/s41524-023-01140-2).

Currently, such correction is activated automatically by the PBC, that now are used to detect whether the corrections should be treated as 1D, 2D or 3D.

A simple example on 2D-hBN is also added to demonstrate the functionality.

Some discussions on the PBC attribute would be needed.


c/dynmat.c
```diff
@@ -567,16 +579,25 @@ static void get_dm(double dm[3][3][2], const long num_patom,
...
         for (j = 0; j < 3; j++) {
             sum += q_cart[i] * dielectric[i][j] * q_cart[j];
         }
     }
+
+    if (dim == 2){
+        q_norm =  2*PI*sqrt(q_cart[0]*q_cart[0] + q_cart[1]*q_cart[1] + q_cart[2]*q_cart[2]);
+        dim_factor = (2./vacuum_size)*q_norm*diel_ext;
+        sum += dim_factor + sum/4.;
+    }
```
atztogo, line 599: Is the new implementation for the calculation, and are the others all interface?

bastonero: Yes, basically this is the core change for calculation. It would be great to have your input for the interface. I also just realized that implemented calculation is not correct - I have to verify and check carefully against QE implementation (probably one also needs an extra parameter). 

I implemented this like 1.5 years ago, and since it relies on PBC I thought to make the PR now. It would be good to have your feedback, happy to change and adapt. Overall it is a quite simple addition, once the units are on spot.

atztogo:
How about followings?
* Remove [phonopy/.nfs00000015ad393e1a00000dd3](https://github.com/phonopy/phonopy/pull/356/files#diff-f5edf7a7754d3afff5cf1c39573ead54eb30ba42c307b0e15b41d9c0b0813569)
* Add tests

I think it is better to put `PBC` and other parameters in `nac_params`, which is found at
https://github.com/phonopy/phonopy/blob/9b58b1c58bc62a71b9bcdc49e0ccdd71062f9cb3/phonopy/api_phonopy.py#L737-L752

PR #358, opened 2024-05-11
support non-collinear magmom add for PWmat caculator

Liang-jianxin:
Adapted to new version

 The latest phonopy  seems  not to support generating magnetic of  supercell from magnetic information in unit cells

here improve a function  that writing of magnetic moments and non-collinear magnetic moments form Tag MAGMOM for pwmat


atztogo:
Is this ready for merging?

PR #359, opened 2024-05-13
fix a bug

Liang-jianxin:
Firstly, as a beginner, thank you very much for your guidance and help in the previous PR.
Here， an error has been changed ,
Conflict between reading magnetic moments from structure and tag MAGMOM, was wrong and not used.


atztogo:
The commit [c817ac8](https://github.com/phonopy/phonopy/pull/359/commits/c817ac866d797f7105f3d59d56c7e784b85c064b) by pre-commit.ci service may indicate that you could run pre-commit before your commit in order to format the code.

PR #360, opened 2024-05-24
Allow for `pip install phonopy[plot]` to get plotting dependencies

Andrew-S-Rosen:
Right now, if one wants to use the band structure plotting utilities, seekpath needs to be manually installed. To help with this and to aid in CI, it would be convenient to support a shorthand of `pip install phonopy[plot]` (or similar) to install the optional dependencies for plotting. This would also allow phonopy to restrict versions of `seekpath` if incompatibilities ever arose.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -251,7 +251,7 @@ def main(build_dir):
             "h5py>=3.0",
             "spglib>=2.3",
         ],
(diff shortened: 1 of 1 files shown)


atztogo:
Thanks!

PR #366, opened 2024-06-13
Add helper functions to go to and from ASE

tomdemeyere:
Two functions to go from and to ASE Atoms object, what do you think?

### DIFF
--- a/phonopy/structure/atoms.py
+++ b/phonopy/structure/atoms.py
@@ -537,6 +537,30 @@ class PhonopyAtoms:
             DeprecationWarning,
         )
         return self.totuple()
+    
+    @classmethod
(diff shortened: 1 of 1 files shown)


atztogo:
I think people will develop their codes that depend on these classmethods. But because I don't use ASE, I can not maintain it well. If ASE changes their API, we have to follow it. At least I am not happy to take care of it.

I made a similar mistake in spgilb (https://github.com/spglib/spglib/pull/386). I decided to drop support of it, but which resulted in making troubles in other codes. PhonopyAtoms is not as influential as spglib, but we don't know the future. I never considered spglib would become so popular.

tomdemeyere:
I see, no worries

I will close this

Issue #380, opened 2024-06-29
QHA API turns all warnings into errors


JaGeo:
This line here https://github.com/phonopy/phonopy/blob/04f912a77e3efbac00a9987b545fc4f5e8927eb9/phonopy/qha/eos.py#L134 turns all warnings into errors. This will affect all subsequent code. 

Could this be made optional in one of the next phonopy versions? I think this is very unsafe once this api is integrated into larger code bases.

(I just spent a longer time debugging to figure out why my code was behaving so strangely)

atztogo:
What is the expected behaviour?

JaGeo:
I would expect that this global warning setting is changed back afterwards

atztogo:
Will inserting `warnings.resetwarnings()` at the last line of `def fit(self, initial_parameters)` fix your issue?

JaGeo:
I will test once I am back. I had used reset to the Default warnings in my code.

JaGeo:
If I use this statement in subsequent code, everything works as expected. Thus, I think inserting the code should work fine. 

I used `warnings.filterwarnings("default")` in my code so far which also seemed to work.

atztogo:
Thank @JaGeo for your checking. I made PR #381. I will merge this if you are fine for this change.

JaGeo:
Thanks! I will close the issue!

PR #381, opened 2024-06-30
Reset warnings for issue #380

atztogo:
Reset warnings for fixing issue #380.

### DIFF
--- a/phonopy/qha/eos.py
+++ b/phonopy/qha/eos.py
@@ -152,3 +152,5 @@ class EOSFit:
             raise
         else:
             self.parameters = result[0]
+
+        warnings.resetwarnings()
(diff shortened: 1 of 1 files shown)


JaGeo:
I've been thinking if the second solution in this answer - using the context module - might be more appropriate as we would otherwise again reset all warning configurations for all vode afterwards. The solution in your PR works in my use case but I am not sure if it works for all potential use cases:
https://stackoverflow.com/a/9134842

I will test the solution mentioned there now.

JaGeo:
So for me, using `with warnings.catch_warnings():` befor changing the warnings filter and move the code afterwards into a block works as well. I think this would be a better solution as it will not change the warning settings users might have set before or afterwards.

atztogo:
Good suggestion. How about 214ba7a?

JaGeo:
This looks great! I would be happy if this could be merged! Thank you so much!

PR #398, opened 2024-07-24
Fix: non-collinear mag missing in supercells generated by using abacus

hongriTianqi:
Dear Prof. Togo,

This is a PR to fix a bug with the abacus interface. We found that in generating supercells with non-collinear magnetism setting, the resulting STRU files lack mag info. This PR is to fix this bug.

This PR can be reviewed by @1041176461 

Thanks.


atztogo:
@1041176461, if you are OK, I will merge.

1041176461:
I think it can be merged.

PR #399, opened 2024-07-25
Fix: collinear case error with abacus interface

hongriTianqi:
Dear Prof. Togo,

Thank you very much for your quick response for my previous PR.

Sorry to bother you again that the modification causes a small error in collinear mag case. This PR makes sure the abacus interface can work for both collinear and noncollinear case now.

Thanks again.
Best

Review Requst to @1041176461


atztogo:
It means probably you have to write a test. Do you want to add the test?

In addition, before to the next commit, Please run pre-commit locally, see https://github.com/phonopy/phonopy?tab=readme-ov-file#pre-commit. You don't need to close this PR, but just from the next commit.

1041176461:
Hi, @hongriTianqi.  I‘ve already added some tests in this file https://github.com/phonopy/phonopy/blob/develop/test/interface/test_abacus.py, may be you can add more for collinear and non-collinear cases.

hongriTianqi:
@atztogo @1041176461 Unit test for non-collinear mag has been added. I see that there exits collinear tests. But why it passed in the previous PR...

test/interface/test_abacus.py
```diff
@@ -61,4 +82,5 @@ def test_read_abacus_output():
 
 if __name__ == "__main__":
     test_read_abacus_mag()
+    test_read_abacus_mag_noncolin()
     test_read_abacus_output()
```
atztogo, line 87: Phonopy employs pytest. You can run your test using pytest. Therefore these lines are unnecessary.

atztogo (approved):

PR #401, opened 2024-08-01
ENH: add script to extract BORN file from QE ph.x output

flokno:
This script is equivalent to `phonopy-vasp-born` but for QE output.

Edit:
The `--primitive_matrix` and `--supercell_matrix` flags are currently not supported


phonopy/interface/qe.py
```diff
@@ -315,7 +330,9 @@ def _set_lattice(self):
 
     def _set_positions(self):
         unit = self._values[0].lower()
-        if "crystal" not in unit:
+        if "angstrom" in unit:
+            self.scaled_positions = False
```
atztogo, line 334: Does this mean you support positions in Cartesian coordinates?

atztogo: If so, could you add a test for parsing such input file because I am not very familiar with qe. The test is `test/interface/test_pwscf.py`.

flokno: Thanks for pointing this out, I fixed an error, added bohr units, and added the tests for the 2 new cases

atztogo:
Thanks @flokno!

atztogo:
I have forgotten to ask about phonopy-vasp-born. How can users use this tool?

flokno:
> I have forgotten to ask about phonopy-vasp-born. How can users use this tool?

You simply run `phonopy-vasp-born` in a folder with `OUTCAR` or `vasprun.xml`. It will write the content of the `BORN` file to screen.

`phonopy-qe-born` works similarly, but one has to specify the input files as explained via the `--help` flag, e.g.

```bash
phonopy-qe-born pw.in ph.out | tee BORN
```

Should we document this somewhere?

atztogo:
Thanks @flokno.

>Should we document this somewhere?

I will take care of it.

PR #409, opened 2024-08-17
Create py.typed

Andrew-S-Rosen:
As recommended in [pep 0561](https://peps.python.org/pep-0561/), a blank `py.typed` marker should be included when type hints are used so downstream codes can type check with `mypy` and similar tools.

There are not too many type hints in phonopy, but there are some!


atztogo:
Thanks @Andrew-S-Rosen!

PR #413, opened 2024-08-21
Address `spglib` dict interface `DeprecationWarnings`

janosh:
```py
spglib.py:115: DeprecationWarning: dict interface (SpglibDataset['international']) is deprecated.Use attribute interface ({self.__class__.__name__}.{key}) instead
spglib.py:115: DeprecationWarning: dict interface (SpglibDataset['transformation_matrix']) is deprecated.Use attribute interface ({self.__class__.__name__}.{key}) instead
...
```

hope i didn't miss any. the amount of `DeprecationWarnings` i'm seeing locally went from 30 to 6 so i may have missed some but without a stacktrace, it's hard to trace them


atztogo:
This requires latest versions of spglib. So I have to think about it.

janosh:
> Why have you changed eigvecs -> eig_vecs?

forgot to revert

atztogo:
Then no problem.

PR #429, opened 2024-09-17
add filename arg to write yaml for mesh and bs

### LINKED ISSUE #336: Phonopy gruneisen api >  no  option to set filename of yaml files
# Issue 

Current implementation of gruneisen api does not provide an option to set file name of yaml file saved for mesh and bandstructures. 

# Solution 

Mimic same code as done in phonopy api that allows user to set file names

I will open an PR with this change in next days

naik-aakash:
Closes #336 

# Changes
Added filename arg to `write_yaml_mesh `and `write_yaml_band_structure ` methods of `PhonopyGruneisen` class. Allows users to modify file names when saving these yaml and hdf5 files when using gruneisen python api

### DIFF
--- a/phonopy/api_gruneisen.py
+++ b/phonopy/api_gruneisen.py
@@ -113,9 +113,9 @@ class PhonopyGruneisen:
         """Write mesh sampling calculation results to file in yaml."""
         self._mesh.write_yaml(filename=filename)
 
-    def write_hdf5_mesh(self):
+    def write_hdf5_mesh(self, filename="gruneisen_mesh.hdf5"):
         """Write mesh sampling calculation results to file in hdf5."""
-        self._mesh.write_hdf5()
+        self._mesh.write_hdf5(filename=filename)
 
     def plot_mesh(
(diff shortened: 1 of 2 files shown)


naik-aakash:
Hi @atztogo, This small change should fix the issue I opened a while ago. I hope it is fine with you as well.

atztogo:
Thanks for your updates.

Could I ask one thing? Could you make the scipy dependency as another PR?

I didn't like to include scipy as a dependency due to potential installation issues in the past. However, if we decide to include it, we need to carefully consider the scipy version. Not everyone will be able to use the latest versions, so leaving the version unspecified should be fine in most cases. Additionally, we’ll need to optimize the GitHub workflow to ensure scipy is installed efficiently. I can handle this aspect, but I suggest adding scipy to the following line:

https://github.com/phonopy/phonopy/blob/1cd8cfd41d517606b8d3d29a48313c9fca74feab/.github/workflows/phonopy-pytest-conda-numpy2.yml#L32

as well as in the other workflows.

PR #434, opened 2024-09-24
Classical statistics

erhart1:
This PR adds support for classical statistics. The functionality is accessible via optional keywords in the python API and CLI.


atztogo:
Thanks @erhart1!

PR #502, opened 2025-01-26
Add chemical formula properties to `PhonopyAtoms`

janosh:
- `formula`: standard chemical formula
- `reduced_formula`: formula with element counts reduced by GCD
- `normalized_formula`: formula with element counts scaled to sum to 1


atztogo:
Thanks @janosh!

PR #508, opened 2025-02-01
Updated phonopy-module.md

Shibu778:
I was going through [Phonopy Python API Documentation](https://phonopy.github.io/phonopy/phonopy-module.html) to understand how to create the `Phonopy` object from already available files during earlier phonon calculations with the Phonopy command line interface. I fixed typo and updated the document to explain the use of `phonopy.cui.load.load` to create `Phonopy` objects.

### DIFF
--- a/doc/phonopy-module.md
+++ b/doc/phonopy-module.md
@@ -146,8 +146,18 @@ displacement_dataset =
 This is set to the `Phonopy` object by:
 
 ```python
-phonopy.dataset = displacement_dataset
+phonon.dataset = displacement_dataset
 ```
+
+Alternatively, if you have `phonopy.yaml` and `FORCE_SETS` file and you want to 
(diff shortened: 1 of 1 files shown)


atztogo:
Thanks for your fix, @Shibu778.

I believe your additional documentation (see below) would be more useful if placed near the bottom of this page, where the current content is a bit outdated. I'll merge your PR and update that section.
``````
Alternatively, if you have `phonopy.yaml` and `FORCE_SETS` file and you want to
create the `Phonopy` object:

```python
from phonopy.cui.load import load
phonon = load("phonopy.yaml", force_sets_filename="FORCE_SETS")
```
It will read the required structure information from `phonopy.yaml` and displacement-force
information from `FORCE_SETS` to make `Phonopy` object.
``````

Issue #439, opened 2024-10-13
Encountered error during pip install phonopy


Shibu778:
I encountered following error while doing `pip install phonopy`. This is happening for a long time now. What is the issue?

```
Building wheels for collected packages: phonopy
  Building wheel for phonopy (pyproject.toml) ... error
  error: subprocess-exited-with-error

  × Building wheel for phonopy (pyproject.toml) did not run successfully.
  │ exit code: 1
  ╰─> [45 lines of output]
      WARNING: Use build.verbose instead of cmake.verbose for scikit-build-core >= 0.10
      2024-10-13 22:38:50,471 - scikit_build_core - INFO - RUN: /tmp/pip-build-env-rux6e5kj/normal/lib/python3.9/site-packages/cmake/data/bin/cmake -E capabilities
      2024-10-13 22:38:50,480 - scikit_build_core - INFO - CMake version: 3.30.4
      *** scikit-build-core 0.10.7 using CMake 3.30.4 (wheel)
      2024-10-13 22:38:50,489 - scikit_build_core - INFO - Build directory: /tmp/tmp77yokri8/build
      *** Configuring CMake...
      2024-10-13 22:38:50,510 - scikit_build_core - INFO - RUN: /tmp/pip-build-env-rux6e5kj/normal/lib/python3.9/site-packages/ninja/data/bin/ninja --version
      2024-10-13 22:38:50,515 - scikit_build_core - INFO - Ninja version: 1.11.1
      2024-10-13 22:38:50,680 - scikit_build_core - INFO - RUN: /tmp/pip-build-env-rux6e5kj/normal/lib/python3.9/site-packages/cmake/data/bin/cmake -S. -B/tmp/tmp77yokri8/build -DCMAKE_BUILD_TYPE:STRING=Release -C/tmp/tmp77yokri8/build/CMakeInit.txt -DCMAKE_INSTALL_PREFIX=/tmp/tmp77yokri8/wheel/platlib -DCMAKE_MAKE_PROGRAM=/tmp/pip-build-env-rux6e5kj/normal/lib/python3.9/site-packages/ninja/data/bin/ninja -DPHONOPY_USE_OMP=ON
      loading initial cache file /tmp/tmp77yokri8/build/CMakeInit.txt
      -- The C compiler identification is GNU 4.8.5
      -- The CXX compiler identification is GNU 4.8.5
      -- Detecting C compiler ABI info
      -- Detecting C compiler ABI info - done
[...]

PR #511, opened 2025-02-04
Add CXX installation error to troubleshooting

ElliottKasoar:
We've had a few people encounter issues installing our package (see: https://github.com/stfc/janus-core/issues/373, similar to #439), which has `phonopy` as a dependency, due to various combinations of `CXX` being unset, or invalid/old compiler versions, which are not uncommon on various HPC systems.

This adds a small section to the installation troubleshooting discussing this, as while the errors tend to be relatively self-explanatory, these errors aren't necessarily particularly familiar to many Python users.

It would be great to be able to point to a section like this, and hopefully it can help other `phonopy` users too.


atztogo:
Merged. Thanks @ElliottKasoar!

PR #522, opened 2025-02-23
Documentation cp2k.rst mod to explain xyz workflow

ladyteam:
As previously discussed. Documentation for SiO2-CP2K-xyz example added.


atztogo:
Thanks @ladyteam!

PR #538, opened 2025-04-13
CP2K forces parser in new 2025 format

ladyteam:
Check of CP2K force files version is based on header and footer strings 'FORCES| Atomic forces', 'FORCES| Total atomic force'. If both persist we consider the calculations were made using CP2K 2025 version and above.
The forces are parsed using 'FORCES|   Atom     x' magic word passed to  iter_collect_forces() function.
The function parse_set_of_forces now check the version and call parse_set_of_forces2025 or parse_set_of_forces2024 depending is the forces are in new or old format. More strings in code but looks safer in case developers of CP2K decide to change force file format again in 2026.


atztogo:
The test failure doesn't come from this PR.

Thanks @ladyteam.

PR #559, opened 2025-05-25
To adapt to the FLEUR MaX-R7.2-9-g430ef20b6 update in reading the FORCE positions.

Youhaojen:
Since the FORCE format in the older version of FLEUR differs from that in the latest MaX-R7.2-9-g430ef20b6 version, the relevant code needs to be updated accordingly.


atztogo:
Sorry for the late response. @Youhaojen, thanks for your PR. I'd like to merge it after addressing the fix proposed by pre-commit. Could you take care of that?

Youhaojen:
@atztogo Thanks for the response. The fix suggested by the pre-commit has been addressed, and the PR has been updated. Please check it. Thank you!

atztogo:
Merged. Thanks @Youhaojen!

atztogo:
@Youhaojen, it seems the FLEUR interface is broken.

https://sourceforge.net/p/phonopy/mailman/message/59295214/

I may fix it, but it is difficult for me because I don't have any input and output files except for the Al example. However, this Al example doesn't work because I guess after this PR.

So I may need
- To fix input structure parser
- To fix the Al example
- Hopefully to have another example with more complex crystal structure

Also ping @neukirchen-212.

atztogo:
@Youhaojen. I used old Al-Fleur example. I confirmed you also updated `FORCES` file, and the current code can parse this file properly. Sorry for my mistake.

But as I wrote above, the current code cannot parse input crystal structure properly. I am fixing it.
If you have another example than Al, I would appreciate it if you could provide it.

PR #561, opened 2025-05-28
Allow necessary info to be passed for file conversion

scohenjanes5:
It is currently impossible to pass the additional information necessary to write many types of input files like qe, wien2k, etc with the file conversion script. I added an optional argument that can accept this information, and wrote clearer error messages explaining why the files can't be built if the info is not provided.

In the case of QE, dummy pp file names were already being written if the names were not provided, but I did not add analogous dummy info for the other calculators that require this extra info because I'm not familiar with their formatting.

### DIFF
--- a/phonopy/interface/calculator.py
+++ b/phonopy/interface/calculator.py
@@ -106,8 +106,10 @@ def convert_crystal_structure(
 ):
     """Convert crystal structures between different calculator interfaces.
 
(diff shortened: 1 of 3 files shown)


scohenjanes5:
Ok, the test passes now that it's not using relative paths to find the POSCAR

PR #591, opened 2025-06-24
Use ThermalDisplacementMatrices to avoid repeat calculations 

quantumsteve:
This use ThermalDisplacementMatrices to calculate the anisotropic Debye–Waller factor once and perform a faster evaluation at each Qpoint. I didn't see anywhere using the temperature value from the private member function `_get_thermal_displacements` so I removed it.

I'm seeing performance improvements of 50x for 40,000 Qpoints.


atztogo:
`ThermalDisplacements` and `ThermalDisplacementMatrices` are written independently, but the latter is enough as you have shown in your PR. I am not sure when I will work for it, but just wrote a small test comparing them at https://github.com/phonopy/phonopy/commit/b835d5ae9e60d051c1a8aa749d14a7d07731b4ad.

PR #592, opened 2025-06-24
Doc tweaks

scohenjanes5:
Fix some typos, reword a few things for clarity, and add calculator interface flags to script examples using calculators other than VASP.


atztogo:
Thanks @scohenjanes5. I have a question about

`phonopy --qe -f disp-001/supercell.out disp-002/supercell.out`

The information about calculator such as `qe` is written in `phonopy_disp.yaml`. In the older `phonopy_disp.yaml` it was not, so the option like `--qe` is necessary, but not right now. However I know these user interface code is fragile and sometimes not working as I expect. And in some example, those calculator name may not be written in `phonopy_disp.yaml` because some examples are old. Did you find any case the following (i.e. no calculator option) doesn't work?

`phonopy -f disp-001/supercell.out disp-002/supercell.out`

Probably what I wrote above should be written in the documentation.

atztogo:
Thanks @scohenjanes5 for your detailed explanation and taking time for describing it. I would like to merge this PR.

I have been developing the phonopy code for long time and initially there was no python API. Use of python language was not very popular in science, and nobody needed the python API. Use of different calculator than VASP was not considered because it was OK if woking for me. Lately, the phonpy code was gradually extended into many directions (e.g. API or supporting different calculators), which was like random construction, for which  I didn't need to consider very much about consistency as the number of users were still few.

I am basically a VASP user and phonopy works fine for the VASP case for the commad-line interface and the python API. I assume majority of users use the command-line interface, I hope that it works intuitively. On the other hand, I haven't considered well about the python API for non-VASP cases. Following your comment, I realized that I will have to start thinking about it, i.e., I have no clear answer to your question (1).

For (2), I agree with your advice. The reason why these two exist is explained below, but when `calculator` is provided, the `factor` should be set following the `calculator`.

The `factor` parameter is historically not only used to unit conversion to THz of phonon frequency, but for any unit such as cm^-1 or meV. In early days, it was not confusing because the code was small and I could take care of each user as the community was small. Probably `factor` should be deprecated, and something like `factor_to_THz` should be prepared, since many features in phonopy relies on it as THz such as thermal properties. In phono3py, the frequency unit is fixed to THz.

PR #594, opened 2025-06-27
Deprecating factor in initiation of Phonopy class

scohenjanes5:
Following the discussion in #592 I modified the `__init__` function of the `Phonopy` class to select the frequency conversion factor only from the selected calculator, preventing situations where the factor and calculator disagree. 

I added a deprecation warning and modified docs and other tools within the package that set the factor command under the following logic: 
If you provide factor explicitly, such as in the settings file, you will get the deprecation error, but if you do not, the code previously used to select the correct factor is removed, since that functionality is done within the `__init__` process.


atztogo:
I introduced `set_facotr_by_calculator` parameter to `Phonopy.__init__`. The default is False not to surprise existing users. By this, I made some small changes at 5cfe3e2. In addition, I wrote some tests at e99496b.

scohenjanes5:
I appreciate your care for the other users!

PR #600, opened 2025-07-04
add reciprocal_lattice and mass to qpoints.hdf5

dmt4:
Would be very useful to have the masses and reciprocal lattice in the new qpoints.hdf5 file. This small change is adding them.


atztogo:
Merged. Thank you, @dmt4.

PR #601, opened 2025-07-07
qlm interface: refactor and preserve input context

dmt4:
This change also relaxes a number of inconvenient restrictions:
    
    * Species names can have suffixes.
    * The input is not restricted to reduced coordinates anymore.
    * Output alat parameter is taken from the input file now.


dmt4:
Nothing urgent. Thanks!

PR #605, opened 2025-07-10
phonon dynamic structure factor: sign error in phase?

quantumsteve:
We observed some odd artifacts when calculating the dynamic structure factor systems like Ge or Si with a non-diagonal primitive matrix. The issue first occurs in phonopy v2.16.0 and we traced the artifacts to changing sign of the phase from
`np.exp(-2j * np.pi * np.dot(pos, G))` to `np.exp(2j * np.pi * np.dot(pos, G_vector))` in #211. 

The [documentation](https://phonopy.github.io/phonopy/dynamic-structure-factor.html) lists the phase as `Q+q` instead of `G=Q-q`. Making this change also eliminates the artifacts and appears more mathematically justified, but also changes the values of some old tests.

Considering the high bar for changing gold data, I'd appreciate the community's advice.

### DIFF
--- a/phonopy/spectrum/dynamic_structure_factor.py
+++ b/phonopy/spectrum/dynamic_structure_factor.py
@@ -228,7 +228,7 @@ class DynamicStructureFactor:
                     G_vector,
                     debye_waller,
                     f,
-                    eigvecs[:, i],
+                    eigvecs[:, i].conj(),
                 )
                 n = bose_einstein_dist(f, self._T)
                 S[i] = abs(F) ** 2 * (n + 1)


atztogo:
I personally received a report that v2.15 worked correctly for Si, but the current version does not. This suggests there may be a bug. Upon brief consideration, I feel your proposed change concerning.

You commented your artifact came from the sign change of the phase. I expect the following gives equivalent result to it (sign change) and this is more reasonable for me.

https://github.com/phonopy/phonopy/blob/4f531c5d951808abe77b1ccd27d31b417b640182/phonopy/spectrum/dynamic_structure_factor.py#L231

Can the change `eigvecs[:, i]` -> `eigvecs[:, i].conj()` give a reasonable result for you case?

quantumsteve:
@atztogo Your suggestion above fixes the artifacts, passes all tests and visually looks better than the output from my branch.  Is there a typo in the documentation?

current develop branch
develop
this branch
with_605
with `eigvecs[:, i] -> eigvecs[:, i].conj()`
with_conj

atztogo:
Great. Thanks for your confirmation.

The documentation is correct. But my implementation was wrong. The definition of q in Squires' famous book differs by a sign from that of phonopy. I followed the equation of the book except for the choice of sign (and eigenvector is different by the definition of dynamical matrix). The equation of $F(Q, q\nu)$ that I wrote in the documentation was confusing. It would be clearer to write it as $F(\mathbf{Q}, \mathbf{q}'\nu)$ (I updated the documentation so). Then $q'=-q$ gives complex conjugate of the eigenvector, and $Q+q' \rightarrow Q-q=G$. In fact, we take summation over q, so q or -q in the first two equations is irrelevant for the calculation, but I wanted to keep the meaning of the scattering.

atztogo:
Thank you very much @quantumsteve. This was a critical bug! Current test doesn't cover. I should add another test.

atztogo:
I added a test with Si.

PR #634, opened 2025-08-19
QpointsPhonon: Always use the C dynamical_matrix_solver

quantumsteve:
One may use `run_dynamical_matrix_solver_c` regardless of whether OpenMP was enabled at build-time.


atztogo:
Thanks @quantumsteve. Previously `phonoc.use_openmp()=True`  was optional, but currently it is True in most cases.

PR #644, opened 2025-09-22
Accelerated version of run_qpoints

PhilipFackler:
This integrates into the existing design a version of `QpointsPhonon.run()` accelerated via numba-cuda and cupy.

1. It will run conditionally if an active device is available (`numba.cuda.is_available()`). This condition could be improved with one that considers the workload size (launching GPU kernels with too small a workload will have worse performance).
2. I wasn't sure about the preferred method of handling dependencies, so I added the minimum. The instructions [here](https://phonopy.github.io/phonopy/install.html#installation-from-source-code) worked for me on two different machines (with and without cuda installed). If you have a GPU that would work for this, you can use something like `conda config --add channels nvidia/label/cuda-12.8.1` and install `cuda-toolkit` from that.
3. Technically this approach will work with rocm/hip as well (since it can "pose" as cuda). I included some code that helps with this from our experiments, but I haven't tested on AMD hardware for this code, and so that would potential require some extra work. (I thought keeping this PR minimal was best.)

### DIFF
--- a/.github/workflows/phonopy-pytest-conda.yml
+++ b/.github/workflows/phonopy-pytest-conda.yml
@@ -23,7 +23,7 @@ jobs:
       run: |
         conda activate test
         conda install --yes -c conda-forge python=${{ matrix.python-version }}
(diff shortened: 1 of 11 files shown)


atztogo:
Considering the future maintenance, what should we do?

PR #650, opened 2025-10-01
update default disp_filename

scohenjanes5:
This function prints a warning if the default kwarg is used, so it is updated to the preferred `phonopy_disp.yaml`


atztogo:
Thanks @scohenjanes5!

Issue #660, opened 2025-10-18
Failure in deploy on conda-forge


atztogo:
@scohenjanes5, `test_conversion.py::test_calc_convert` fails in conda-forge deploy.

```

            "elk": example_dir / "Si-elk/elk.in",
            "crystal": cwd / "Si-CRYSTAL.o",
            "fleur": example_dir / "Al-Fleur/fleur_inpgen",
            "abacus": example_dir / "Al-ABACUS/abacus-pw/STRU.in",
        }
    
        for calc in xtra_inps.keys():
            with tempfile.TemporaryDirectory() as temp_dir:
                vasp_tmp = os.path.join(temp_dir, "vasp_tmp_file")
>               simulate_calc_convert_script(xtra_files[calc], vasp_tmp, calc, "vasp")

test/interface/test_conversion.py:113: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
test/interface/test_conversion.py:72: in simulate_calc_convert_script
    run()
../_test_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehol/lib/python3.13/site-packages/phonopy/scripts/phonopy_calc_convert.py:228: in run
    opts = get_options()
           ^^^^^^^^^^^^^
../_test_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehol/lib/python3.13/site-packages/phonopy/scripts/phonopy_calc_convert.py:93: in get_options
    return parser.parse_args()
           ^^^^^^^^^^^^^^^^^^^
../_test_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehol/lib/python3.13/argparse.py:1898: in parse_args
    args, argv = self.parse_known_args(args, namespace)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #661, opened 2025-10-18
Release 2.43.5 and fix test for conda-forge deployment

atztogo:
For #660.
Probably relative path to example directory is not a good idea.


scohenjanes5:
Just seeing the two issues today. Glad it was a simple fix. I'll keep that in mind for the future.

PR #778, opened 2026-03-25
Support hdf5 saving

thomasloux:
In the current code, Phonopy.save proposes a hdf5_setting but no implementation was available.

Yaml file is not super efficient for large scale saving of Phonopy objects, at least in my experience. I've added hdf5 support for that. I need to check whether is it indeed faster to load.

In my test, the loaded Phonopy is identical, although the hdf5 does not save exactly the same information. Some information in yaml are not used when loaded (symprec / symmetry_tolerance for instance).

### DIFF
--- a/phonopy/api_phonopy.py
+++ b/phonopy/api_phonopy.py
@@ -3146,15 +3146,19 @@ class Phonopy:
         self,
         filename: str | os.PathLike = "phonopy_params.yaml",
         settings: dict | None = None,
-        hdf5_settings: dict | None = None,
         compression: str | bool = False,
     ) -> str:
         """Save phonopy parameters into file.
 
+        The output format is determined by the filename extension:
+        - ``.yaml`` or ``.yml``: YAML format (optionally xz-compressed).
+        - ``.hdf5``: Self-contained HDF5 file with structure metadata
+          embedded as YAML and force constants / force sets as binary datasets.
+
         Parameters
         ----------
         filename: str, optional
-            File name. Default is "phonopy_params.yaml"
+            File name. Default is "phonopy_params.yaml".
         settings: dict, optional
             It is described which parameters are written out. Only the settings
             expected to be updated from the following default settings are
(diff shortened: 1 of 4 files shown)


atztogo:
I'm not a big fan of HDF5, and I'd like to minimize the maintenance burden of the phonopy codebase, since I'm essentially the sole maintainer supporting a large number of users. It's a significant responsibility. My thinking is that most small data can be stored in YAML, and only large datasets need to be saved in HDF5. If you think this strategy is reasonable, could you let me know which data would qualify as 'large'?

thomasloux:
I need to check but when I tested some high throughput phonon calculation, it was faster to recompute the phonon than reading the force sets on the disk to build the force constant. I can evaluate that for both yaml and hdf5 and increasing system size yes. Note: I have some fast H200 gpus.

thomasloux:
I've realized what takes a really long time is the symmetrization of the force constant using use_symfc_projector=True. I'll give concrete figures, but parsing the yaml file can be extremely slow as well, although this mostly applies in my case for a system with 54 atoms and supercell 4x4x4: 30 seconds to read yaml file without building the force constants and 1 sec for hdf5 file.

atztogo:
I'm not entirely sure what issue you're experiencing. There may be another way to resolve it within the current implementation using the appropriate settings. For this reason, I'd prefer not to merge this PR.

thomasloux:
[benchmark_yaml_vs_hdf5.ipynb](https://github.com/user-attachments/files/26411431/benchmark_yaml_vs_hdf5.ipynb)

Attaching a benchmark that checks:
- loading force sets
- loading force sets + construction of force constant
- loading force sets + construction of force constant (FC) + FC symmetrization
- File sizes

I test:
- a small NaCl system
- a synthetic bigger system with unit cells with 16 or 54 atoms
- a real system with 24 atom unit cell: (https://next-gen.materialsproject.org/materials/mp-567878)

TLDR: 
- HDF5 is way faster to read force sets compared to yaml (often at least 10x factor)
- In a realist setup where a user read a file and get a symmetrized FC, the difference is only <20% (hdf5 faster), because symmetrization takes a lot of time. The difference only comes from reading force sets, so it's negligible for small systems. 
- HDF5 is way smaller (factor 3 at least) compared to yaml, but compressed yaml (with xy, as provided in phonopy) is always of the same size, or slightly smaller, when only force sets are present. With FC in the file, the file is significantly smaller compared to hdf5.

Loading time
image

File size for the same system
image

For larger unit cell (and supercell 2x2x2). Loading time of yaml increases quite fast with file size (parsing of string)
image

Conclusion: I agree that hdf5 is no really needed for system with < 100 atoms in unit cell, because compressed yaml is light and because in the end, the bottleneck is the symmetrization of the force constant. Hdf5 is still way better in the case of loading phonopy file where a force constant is saved. But because force constant is really large, I suspect this case is pretty rare.

atztogo:
Thanks for your report. Basically I prefer text format, because it is robust. The following is my comment.

- For fast reading of force sets, you can use `FORCE_SETS` file.
- For symmetrization, `phonopy-load --fc-calc traditional` or `phonopy --fc-calc traditional --fc-symmetry` for type-1 force sets, https://phonopy.github.io/phonopy/input-files.html#force-file-force-sets. Use of symfc (per default) is more accurate as the symmetrization, but `--fc-calc traditional` is good enough in most cases. For type-2 force sets, `--fc-calc traditional` cannot be used.
- For file size of `FORCE_SETS` can be a matter, I can support reading `FORCE_SETS.xz`. In fact, I thought I had made it so, but it seems it wasn't.
- You can store force constants in `force_constants.hdf5`, too. I can support reading `FORCE_CONSTANTS.xz` if requested.

PR #799, opened 2026-05-06
Fix copy-by-reference in qpoints.py

julian-saure:
Due to the copy-by-reference on line 279 of qpoints.py, calculating eigenvalues, eigenvectors and dynamical matrices using run_qpoints() always returns incorrect dynamical matrices equal to the returned eigenvectors.

Adding .copy() will instead copy by value and the dynamical matrices should be returned correctly.


atztogo:
@julian-saure, thanks for pointing that out. I modified your commit while preserving the meaning but saving memory in case eigenvectors are returned — though on modern machines, the difference should be negligible. I also added a test.

atztogo:
Merged. @julian-saure, in case my modification is wrong, please reopen and give me a comment.

PR #858, opened 2026-06-09
Fix a numpy bug in EOS fit

Andrew-S-Rosen:
This PR fixes a bug in the `vinet` EOS fit related to how numpy handles fractional exponents.

The code below is the current `vinet` fit. As currently implemented, raising a negative number to a non-integer exponent is treated as invalid and so making the `x` variable can fail, particularly when running calculations at high pressures. Using `np.cbrt` (cube-root) instead of `** (1.0 / 3)` fixes this by returning the real-valued cube-root.

```python
    def vinet(v: NDArray[np.double], p: NDArray[np.double]) -> NDArray[np.double]:
        """Return Vinet EOS.

        p[0] = E_0
        p[1] = B_0
        p[2] = B'_0
        p[3] = V_0

        """
        x = (v / p[3]) ** (1.0 / 3)
        xi = 3.0 / 2 * (p[2] - 1)
        return p[0] + (
            9 * p[1] * p[3] / (xi**2) * (1 + (xi * (1 - x) - 1) * np.exp(xi * (1 - x)))
        )
```

For example, the following was what was erroring out for me when trying to create `x`. But switching to `np.cbrt` resolves it.

```python
import numpy as np

x = (
    np.array(
        [
            46.191144,
            53.889668,
            61.58819199,
            69.28671599,
            73.13597799,
            73.90583039,
            74.67568279,
            75.44553519,
            76.21538759,
            76.98523999,
            77.75509239,
            78.52494479,
            79.29479719,
            80.06464959,
            80.83450199,
            84.68376399,
            92.38228799,
        ]
    )
    / np.float64(-0.18013697985445276)
) ** (1.0 / 3)
```

### DIFF
--- a/phonopy/qha/eos.py
+++ b/phonopy/qha/eos.py
@@ -88,7 +88,7 @@ def get_eos(eos: str) -> EosFunc:
         p[3] = V_0
 
         """
-        x = (v / p[3]) ** (1.0 / 3)
+        x = np.cbrt(v / p[3])
         xi = 3.0 / 2 * (p[2] - 1)
         return p[0] + (
             9 * p[1] * p[3] / (xi**2) * (1 + (xi * (1 - x) - 1) * np.exp(xi * (1 - x)))
--- a/test/qha/test_eos.py
+++ b/test/qha/test_eos.py
@@ -1,0 +1,112 @@
+"""Tests for equation-of-state functions in phonopy.qha.eos."""
+
+from __future__ import annotations
+
+import numpy as np
+import pytest
+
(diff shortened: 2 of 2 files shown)


atztogo:
Can I use the data above to write a test?

Andrew-S-Rosen:
@atztogo: Please go ahead! I cannot tell you how many hours it took me to figure this out! 😭

atztogo:
Thanks @Andrew-S-Rosen!

Andrew-S-Rosen:
Thanks for your tests and quick review!

Issue #859, opened 2026-06-09
Additional clarity on units


Andrew-S-Rosen:
There are a few places throughout phonopy where additional clarity on units would be helpful:
- `qha.plot_gibbs_temperature()` does not show units on the y-axis
- For the various property attribute docstrings, it would be helpful to include them in the docstring. e.g. [here](https://github.com/phonopy/phonopy/blob/faaa9a49222252195acb9f2d3fd8679b4e6889cb/phonopy/api_qha.py#L190-L191)

Part of the reason for this is that the thermal property YAML files have free energies in kJ/mol, but elsewhere other free energies are in eV. It would be useful to have the units annotated for `thermal_properties.free_energies`, `qha.gibbs_temperature`, etc. to prevent mistakes. In general, all plots would benefit from having units labeled, and all properties would benefit from having units in the docstring.

atztogo:
True.

mvanhorn:
Opened #880 addressing the QHA portion of this: the `PhonopyQHA.plot_gibbs_temperature` / `plot_bulk_modulus_temperature` wrappers were passing unit-less `ylabel` defaults that overrode the unit-bearing defaults already in `qha/core.py`, so the default plots now read "Gibbs free energy (eV)" and "Bulk modulus (GPa)". I also annotated the units in the corresponding `PhonopyQHA` property docstrings (gibbs_temperature eV, bulk_modulus_temperature GPa, volume_temperature Angstrom^3, thermal_expansion K^-1, helmholtz_volume eV) and added a regression test. The broader sweep across all plots and properties is left for follow-up.

mvanhorn:
Submitted https://github.com/phonopy/phonopy/pull/886 for the docstring-units half of this issue (the QHA ylabel half already landed in #860) - it adds physical units (K, kJ/mol, J/K/mol) to the `get_thermal_properties_dict` docstring, with a test that cross-checks them against the `ThermalProperties` source so they can't drift.

PR #880, opened 2026-06-25
docs: include units in QHA plot y-axis labels

mvanhorn:
## Summary

Annotate the QHA plot y-axis labels with their physical units so the default plots are self-describing, and add the missing units to the corresponding `PhonopyQHA` property docstrings.

## Why this matters

Issue #859 asks for clearer unit annotations across phonopy, calling out two concrete QHA pain points: `qha.plot_gibbs_temperature()` renders a y-axis with no units, and the property attribute docstrings omit the units of the returned arrays. Because thermal-property energies appear in different units across phonopy's I/O (eV vs kJ/mol), an unlabeled axis or docstring makes unit-mismatch mistakes easy.

The public `PhonopyQHA.plot_gibbs_temperature` / `plot_bulk_modulus_temperature` wrappers passed unit-less `ylabel` defaults (`"Gibbs free energy"`, `"Bulk modulus"`) that overrode the unit-bearing defaults already present in `phonopy/qha/core.py`. This change makes the wrappers consistent with `core.py`.

## Changes

- `phonopy/api_qha.py`
  - `plot_gibbs_temperature`: default `ylabel` is now `"Gibbs free energy (eV)"`.
  - `plot_bulk_modulus_temperature`: default `ylabel` is now `"Bulk modulus (GPa)"`.
  - Added units to the returned-array docstrings of `gibbs_temperature` (eV), `bulk_modulus_temperature` (GPa), `volume_temperature` (Angstrom^3), `thermal_expansion` (K^-1), and `helmholtz_volume` (eV).
- `test/qha/test_QHA.py`
  - Added `test_qha_plot_units`, which builds the existing `qha_si` fixture, asserts the Gibbs and bulk-modulus plot y-labels carry `(eV)` / `(GPa)`, and guards the unit annotations in the property docstrings.

This is a labels/docstrings-only change; no numeric behavior is altered.

## Testing

`pytest test/qha/test_QHA.py` (Agg backend) -> 11 passed, including the new `test_qha_plot_units` and all existing numeric QHA regression assertions, confirming the change is label/docstring-only.

This addresses the QHA portion of the issue; the broader sweep across all plots and properties is left for follow-up.

Refs #859

### DIFF
--- a/phonopy/api_qha.py
+++ b/phonopy/api_qha.py
@@ -154,7 +154,7 @@ class PhonopyQHA:
         Returns
         -------
         list
-            Thermal expansion coefficients at temperatues.
+            Thermal expansion coefficients at temperatues in K^-1.
             shape=(temperatures, )
 
         """
@@ -167,7 +167,7 @@ class PhonopyQHA:
         Returns
         -------
         ndarray
-            Helmholtz free energies calculated at temperatures and volumes.
+            Helmholtz free energies calculated at temperatures and volumes in eV.
             shape=(temperatures, volumes)
 
         """
@@ -180,7 +180,7 @@ class PhonopyQHA:
         Returns
         -------
         ndarray
-            Equilibrium volumes at temperatures
+            Equilibrium volumes at temperatures in Angstrom^3.
             shape=(temperatures,), dtype=float
 
         """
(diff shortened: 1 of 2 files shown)


atztogo:
Thanks @mvanhorn!

PR #886, opened 2026-06-26
docs: add physical units to thermal properties dict docstring

mvanhorn:
## Summary

`Phonopy.get_thermal_properties_dict` documented each returned array only by shape and dtype, not by physical unit, so a caller reading the docstring could not tell whether `free_energy` was in eV, kJ/mol, or something else. This adds the physical units to the docstring for the four returned quantities, matching the units the underlying `ThermalProperties` arrays are actually computed in.

## Why this matters

Issue #859 asks for the thermal-properties units to be discoverable from the API. Without them, users have to read the source of `ThermalProperties` to find that temperatures are in K and the energy/entropy/heat-capacity quantities are per-mole. Stating the units where the values are returned removes that round-trip and the risk of unit errors in downstream analysis.

The units added are taken from the `ThermalProperties` source, not invented: temperatures in K, Helmholtz free energy in kJ/mol, entropy in J/K/mol, and heat capacity in J/K/mol.

## Changes

`phonopy/api_phonopy.py` adds a one-line unit description under each of `temperatures`, `free_energy`, `entropy`, and `heat_capacity` in the `get_thermal_properties_dict` docstring. No behavior changes.

## Testing

Added `test_thermal_properties_dict_docstring_units` in `test/api/test_api_phonopy.py`. It asserts the docstring carries the expected unit text and, to keep the documentation honest, cross-checks each unit against the `ThermalProperties` source docstring so the doc cannot drift from the actual units. `pytest test/api/test_api_phonopy.py` passes (17 passed, 2 skipped).

Refs #859

### DIFF
--- a/phonopy/api_phonopy.py
+++ b/phonopy/api_phonopy.py
@@ -2514,13 +2514,17 @@ class Phonopy:
         'free_energy', 'entropy', and 'heat_capacity'.
         Each value of corresponding key is as follows:
 
-        temperatures: ndarray
+        temperatures : ndarray
+            Temperatures in K.
             shape=(temperatures, ), dtype='double'
         free_energy : ndarray
+            Helmholtz free energies in kJ/mol.
             shape=(temperatures, ), dtype='double'
         entropy : ndarray
+            Entropies in J/K/mol.
             shape=(temperatures, ), dtype='double'
         heat_capacity : ndarray
+            Heat capacities in J/K/mol.
             shape=(temperatures, ), dtype='double'
 
         .. deprecated::
--- a/test/api/test_api_phonopy.py
+++ b/test/api/test_api_phonopy.py
@@ -1,6 +1,7 @@
 """Tests of Phonopy API."""
 
 import copy
+import inspect
 from pathlib import Path
 
 import numpy as np
@@ -9,6 +10,7 @@ import pytest
 import phonopy
 from phonopy import Phonopy
 from phonopy.interface.pypolymlp import PypolymlpParams
+from phonopy.phonon.thermal_properties import ThermalProperties
 from phonopy.structure.cells import isclose
 from phonopy.structure.dataset import get_displacements_and_forces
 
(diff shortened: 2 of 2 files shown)


atztogo:
Thanks @mvanhorn. Is a better systematic approach of physical units needed for human agents, coding agents, or both?

Andrew-S-Rosen:
@atztogo I do think some sort of additional clarity could be helpful. It can get confusing rather quickly. For instance, some parts throughout phonopy return the phonon contribution to the (e.g. Helmholtz) free energy, whereas other parts return the total system (e.g. Gibbs) free energy. Some parts throughout the code use eV (e.g. as with Gibbs free energy), whereas [other parts use kJ/mol](https://github.com/phonopy/phonopy/blob/a70f40911607bd30f3f034462cc2cf7de802ee7a/doc/examples.md?plain=1#L144). Then there are potential ambiguities with respect to what "mol" is specifically referring to --- is it the mol of formula units, and how are those formula units defined; for instance, are the formula units with respect to the input system or any automatically generated primitive cell (which becomes more complicated now that primitive matrices are constructed by default)? For new users and I imagine likely for agents as well, this is very easy to mix up. This was my motivation beyond #859.

Beyond even clearer docstrings and documentation, I am not sure if there is a better systematic approach that also does not break backwards compatibility.

atztogo:
This PR is focused on the `Phonopy` class, but I appreciate that your advice points to something broader. I think this is largely a design issue, and we may need to rethink the division of roles between the `Phonopy` and `PhonopyQHA` classes. I see some people want to improve `PhonopyGruneisen`.

As background, I never really intended for users to use `PhonopyQHA` or `PhonopyGruneisen`, or more precisely, I'd rather they didn't. This is because I'm aware that these were implemented at the "good enough to just work" level. If users actually need them, I think we'd have to sit down and rethink them properly. I had also considered the possibility that someone else might build something better and distribute it. However, what I've learned from my long experience as a developer is that maintaining code is harder than writing it. In that sense, I'm starting to think it might be worth having something corresponding to `PhonopyQHA` and `PhonopyGruneisen` a bit better in the phonopy project.

Andrew-S-Rosen:
Noted about the focus on `Phonopy` here --- I had not noticed. 

I think a lot of people are using (and getting significant value from) the `PhonopyQHA` class. I thinks this is a testament to the hard work you have already done to get it to this stage. I do not know that there is a better alternative out there --- at least, not one I am happy with!

PR #988, opened 2026-09-12
Add QHA system entropy and enthalpy at constant pressure

### LINKED ISSUE #889: Feature request: provide an enthalpy and entropy property in the `PhonopyQHA`
The `PhonopyQHA` class contains many important thermodynamic properties, such as `gibbs_temperature`, which is G(T, P) at the user-specified pressure and range of temperatures. It would be nice to also return the enthalpy and entropy at the user-specified pressure and range of temperatures (e.g. an analogous `enthalpy_temperature` and `entropy_temperature`). Like with the Gibbs free energy, I am referring to the enthalpy and entropy of the system, not just the phonon contribution. 

Since (dG/dT)_P = -S, it should be a matter of just taking the numerical gradient of G with respect to T and negating the result, provided that the user's temperature grid is sufficiently fine. This would give you S at the user-specified pressure and across the range of temperatures.

Then from G = H - TS, once entropy is obtained, it should be trivial to get the enthalpy as well.

There are most certainly other ways to do this, e.g. by starting from the computed equilibrium volume and using the `entropy-volume` data that is returned. I'm not sure which is the most efficient from an API perspective, but the approach I outlined above seems rather straightforward and also ensures everything is in the same units as the returned Gibbs free energies.

shaneraphel:
## Summary

Closes #889.

`PhonopyQHA` already returns `gibbs_temperature` = G(T, p). This adds the matching system (not phonon-only) values

- `entropy_temperature` in J/K/mol
- `enthalpy_temperature` in eV

and the same fields on `QHAResult` from `run_qha`.

S is **not** obtained from a numerical (−dG/dT)_p. Following the comment on #889, S(T, p) = S(T, V_eq(T, p)) using the same degree-4 S(V) polynomial already used by `heat_capacity_P_polyfit`. Then H = G + TS (S converted eV/K so the identity holds in the same units as G).

On the deprecated `PhonopyQHA` path this is refused when temperature-dependent electronic free-energy arrays are supplied, because S_el is not stored there. `run_qha(..., electronic_structures=...)` already adds S_el to the entropy grid, so the new `QHAResult` fields include the electronic contribution.

Writers: `write_entropy_temperature` / `write_enthalpy_temperature` (legacy class and `phonopy.qha.output`).

## Checks

- Exact quadratic S(V) is recovered by the degree-4 fit (`test_entropy_enthalpy.py`).
- H = G + TS holds to 1e-12 eV on the Si QHA fixture.
- `run_qha` and `PhonopyQHA` agree bitwise on NaCl (with and without p = 5 GPa).
- Spot-check only (not a CI bound): on Si, max relative |S − (−dG/dT)| is 2.7×10⁻³ for T > 100 K (10 K grid). That residual is the reason the implementation follows S(V_eq) rather than differentiating G.

This is a thermodynamic-API completion, not a materials discovery.

## Test plan

- [x] `pytest test/qha/test_entropy_enthalpy.py test/qha/test_QHA.py test/qha/test_QHA_efe.py test/qha/test_qha_output.py test/qha/test_run_qha.py::test_run_qha_matches_phonopy_qha test/qha/test_run_qha.py::test_run_qha_electronic_structures`
- [x] ruff on the touched files

Made with [Cursor](https://cursor.com)

### DIFF
--- a/phonopy/qha/calc.py
+++ b/phonopy/qha/calc.py
@@ -88,6 +88,74 @@ def compute_heat_capacity_p_numerical(
     return np.array(cp, dtype="double")
 
 
+class EntropyEnthalpyArrays(NamedTuple):
+    """System entropy and enthalpy at constant pressure.
+
+    Both arrays have the same length as the input temperatures. Entropy is
+    in J/K/mol and enthalpy is in eV, so that G = H - T S holds after
+    converting S to eV/K.
+
+    """
+
+    entropy: NDArray[np.double]
+    enthalpy: NDArray[np.double]
+
+
+def compute_entropy_enthalpy_temperature(
+    temperatures: NDArray[np.double],
+    volumes: NDArray[np.double],
+    equilibrium_volumes: NDArray[np.double],
+    entropy: NDArray[np.double],
+    gibbs_free_energies: NDArray[np.double],
+) -> EntropyEnthalpyArrays:
+    """Evaluate S(T, p) and H(T, p) from S(V) fits at V_eq.
+
+    S(T, p) = S(T, V_eq(T, p)) where S(V) is a degree-4 polynomial at each
+    temperature, the same interpolation used for heat_capacity_P_polyfit.
+    G(T) is not differentiated. H = G + T S with S converted eV/K.
+
+    Parameters
+    ----------
+    temperatures : ndarray
+        Temperatures in K. shape=(num_elems,)
+    volumes : ndarray
+        Unit cell volumes of the input volume grid in angstrom^3.
+        shape=(volumes,)
+    equilibrium_volumes : ndarray
+        Equilibrium volumes at temperatures in angstrom^3.
+        shape=(num_elems,)
+    entropy : ndarray
+        Entropies at constant volume in J/K/mol, indexed consistently
+        with temperatures. shape=(>=num_elems, volumes)
+    gibbs_free_energies : ndarray
+        Gibbs free energies at temperatures in eV. shape=(num_elems,)
+
+    """
+    if len(volumes) < 5:
+        raise RuntimeError(
+            "At least 5 volume points are needed to fit S(V) to a "
+            "polynomial of degree 4."
+        )
+
+    ev_to_jmol = get_physical_units().EvTokJmol * 1000.0
+    n = len(temperatures)
+    entropies = np.empty(n, dtype="double")
+    for i in range(n):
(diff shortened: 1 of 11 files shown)


phonopy/qha/calc.py
```diff
@@ -88,6 +88,74 @@ def compute_heat_capacity_p_numerical(
...
 
 
+class EntropyEnthalpyArrays(NamedTuple):
+    """System entropy and enthalpy at constant pressure.
+
+    Both arrays have the same length as the input temperatures. Entropy is
+    in J/K/mol and enthalpy is in eV, so that G = H - T S holds after
+    converting S to eV/K.
+
+    """
```
atztogo, line 98: I prefer dataclasses over NamedTuples unless there's a specific reason to use a NamedTuple. But I see I used it by myself. I will refactor it later by myself.

phonopy/qha/calc.py
```diff
@@ -88,6 +88,74 @@ def compute_heat_capacity_p_numerical(
...
+            "polynomial of degree 4."
+        )
+
+    ev_to_jmol = get_physical_units().EvTokJmol * 1000.0
+    n = len(temperatures)
+    entropies = np.empty(n, dtype="double")
+    for i in range(n):
+        try:
+            parameters = np.polyfit(volumes, entropy[i], 4)
+        except np.lib.polynomial.RankWarning as exc:  # type: ignore
```
atztogo, line 146: I will address the following by myself after merging,
`from numpy.exceptions import RankWarning`

atztogo (approved):

atztogo:
Thanks @shaneraphel.
My implementation is not nice because physical unit system is not uniform.

atztogo:
@shaneraphel, in #990, I changed the physical unit system in `run_qha` and `phonopy/qha/calc.py`. Both the `PhonopyQHA` and `QHA` classes still use the same unit system.

```
Three of these replacements change units. `QHAResult.entropy_temperature`
and `QHAResult.heat_capacity_P.heat_capacities` are in eV/K where
`PhonopyQHA` reported J/K/mol, and `QHAResult.bulk_moduli` is in
eV/angstrom^3 where `bulk_modulus_temperature` reported GPa. `QHAResult` is
in eV, angstrom and K throughout, and the writers in `phonopy.qha.output`
convert back, so the output files are unchanged.
```

shaneraphel:
tanks for review!

